PR #299, opened 2023-06-29
Fixed inclusion of image tools controlled by variables OMNECT_TOOLS and OMNECT_DEVEL_TOOLS

HarryWaschkeit:


recipes-core/os-release/os-release.bbappend
```diff
@@ -1,5 +1,7 @@
 inherit metadata_scm
 
+require recipes-omnect/images/omnect-os-tools.inc
```
mlilien, line 3: use a class or handle setting OMNECT_TOOLS and OMNECT_OS_TOOLS in distro.conf (i prefer that)
imho it's not a good idea to include a different recipe group path

HarryWaschkeit: I'm not sure whether it is possible to put that in to os-distro.conf.
that was my first thought, too, but the python function couldn't be parsed (iirc bitbake uses a first pass to gather configuration files where no full parsing is yet active).
I put it into a class.

mlilien (approved):

PR #300, opened 2023-06-30
General cellular (LTE) support by example rpi4-omnect-lab

HarryWaschkeit:
pull request got replaced by #365 

for convenient reference a link to the former pr stemming from my private fork #283 

Before a merge can be done the rootfs size topic still needs to be cleared.


conf/distro/include/omnect-os-kernel.conf
```diff
@@ -6,7 +6,10 @@ OMNECT_KERNEL_SRC_URI = " \
     file://disable-kvm.cfg \
     file://docker.cfg \
     file://lxc.cfg \
+    file://mbim.cfg \
     file://netfilter.cfg \
+    file://qualcomm.cfg \
+    file://sierra.cfg \
```
mlilien, line 12: can we add this depended on "3g"?

conf/machine/rpi4-test.extra.conf
```diff
@@ -0,0 +1,21 @@
+require conf/machine/include/rpi.inc
```
mlilien, line 1: why do we need both rpi-test.conf and rpi-test.extra.conf ?

conf/machine/rpi4-test.extra.conf
```diff
@@ -0,0 +1,21 @@
...
+
+# wic: how to partition
+WKS_FILE = "omnect-os.rpi4.wks.in"
+
+# u-boot boot cmd config vars
+UBOOT_ARCH = "arm64"
+
+OMNECT_PART_SIZE_ROOTFS    ?= "900000"
+
+OMNECT_ONLINE_INTERFACES = "${OMNECT_ETH0}"
```
mlilien, line 18: why not WWAN0 resp. WLAN0?

dynamic-layers/raspberrypi/recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -3,9 +3,12 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 SRC_URI += "\
     file://add-reset-info.patch \
     file://enable-reset-info-cmd-fragment.cfg \
+    file://fix-usb-scanning-upstream.patch \
+    file://fix-usb-scanning-balena.patch \
     file://omnect_env.patch \
     file://rpi-always-set-fdt_addr-with-firmware-provided-FDT-address.patch \
     file://omnect_env_rpi.h \
+    file://0001-usb-xhci-ring.c-don-t-queue-URB-for-halted-endpoint.patch \
```
mlilien, line 11: why do we need this patches?
how do we plan on getting rid of them?
does upstream plan to integrate them?
please add comments.

can we add them depending on distro feature "3g"?

recipes-connectivity/networkmanager/networkmanager_%.bbappend
```diff
@@ -0,0 +1,15 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI += " \
+	file://0001-wwan-fix-provision-of-IPv6-addresses.patch \
```
mlilien, line 4: is upstream aware of this problematic? is there a plan to integrate this patch?
is this really a fix? as far as i understand the patch disables ipv6?

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -23,8 +23,8 @@ do_install:append() {
         install -m 0644 ${WORKDIR}/80-wlan.network ${D}${systemd_unitdir}/network
         sed -i 's/^Name=wlan0/Name=${OMNECT_WLAN0}/' ${D}${systemd_unitdir}/network/80-wlan.network
         # configure systemd-networkd-wait-online success if any of eth0 or wlan0 are online
-        sed -i -e 's#^ExecStart=\(.*\)#ExecStart=/bin/bash -c \x27\1 --any --interface=${OMNECT_ETH0} --interface=${OMNECT_WLAN0} --timeout=\${OMNECT_WAIT_ONLINE_TIMEOUT_IN_SECS:-300}\x27#' \
-            ${D}${systemd_system_unitdir}/systemd-networkd-wait-online.service
+        #sed -i -e 's#^ExecStart=\(.*\)#ExecStart=/bin/bash -c \x27\1 --any --interface=${OMNECT_ETH0} --interface=${OMNECT_WLAN0} --timeout=\${OMNECT_WAIT_ONLINE_TIMEOUT_IN_SECS:-300}\x27#' \
+        #    ${D}${systemd_system_unitdir}/systemd-networkd-wait-online.service
```
mlilien, line 27: dead code

recipes-connectivity/networkmanager/networkmanager_%.bbappend
HarryWaschkeit: yes, the patch is really necessary if we want to use IPv6 addresses as propagated by broadband provider, I tested this.
I found it in the pull request https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1170 which didn't make it into the code at first.
However, beginning from version 1.38 this patch is included as of commit https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/586afdfc5298285d09a5b4965690899833a21abb (part of version 1.37.3-dev-52-g586afdfc52)

recipes-core/systemd/systemd_%.bbappend
HarryWaschkeit: removed

recipes-connectivity/modemmanager/modemmanager_%.bbappend
```diff
@@ -0,0 +1,9 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+do_install:append() {
+    # FCC unlock for Sierra Wireless EM7455
+    install -d ${D}${sysconfdir}/ModemManager/fcc-unlock.d/
```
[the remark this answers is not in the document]
HarryWaschkeit: you are right, this is something to inject in images as needed and hence belongs to omnect-os.
removed it completely for the time being, but I can bring it back - commented out plus some additional hints - if we decide to keep it for documentation purposes.

JanZachmann (changes requested):

conf/machine/rpi4-omnect-lab.conf
```diff
@@ -0,0 +1,26 @@
...
+require conf/machine/raspberrypi4-64.conf
+require conf/machine/raspberrypi4-64.tpm.conf
+
+require conf/machine/include/rpi.inc
+
+DISTRO_FEATURES:append = " 3g tpm2"
+MACHINE_FEATURES:append = " 3g tpm2"
+
+# adu compatibility id
+ADU_DEVICEPROPERTIES_COMPATIBILITY_ID = "2"
```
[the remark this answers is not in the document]
HarryWaschkeit: good point, changed

JanZachmann (approved):

dynamic-layers/raspberrypi/recipes-bsp/u-boot/u-boot_%.bbappend
mlilien: we deciced to deactivate usb for now

recipes-core/systemd/systemd/aziot-identityd-startup-timeout-lte.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+TimeoutStartSec=180
```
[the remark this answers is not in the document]
HarryWaschkeit: I don't have problems with setting this value globally, but I refrained from doing so, because other tests are obviously running fine with the much smaller timeout.
I know however that with the standard timeout of 60s (systemd default) the smoke test failed very often, even 120s weren't enough what perplexed me too to some extent, but apparently the identityd had several more attempts until it succeeded to fully start up.
in any case that meant to me (and still does) that this is specific to LTE for whatever reason.

conf/machine/rpi4-test.extra.conf
HarryWaschkeit: yes, in the end somewhere has to be defined which interfaces are available for each interface class, and for Ethernet this can vary between one and three interfaces now.
this depends on the target only, at least at the moment, it might change if we supported several models of the same hardware with varying number of Ethernet interfaces.

of course I could change that to always assume OMNECT_ETH0 being set, so that in omnect-os-distro.conf it can be unconditionally added.
and if you find it more elegant, the same could be also done there depending on OMNECT_ETH[12...] being defined.

mlilien: i would prefer if OMNECT_ONLINE_INTERFACES is not semi-"dynamic" and i can see its value in the corresponding machine config.

recipes-core/systemd/systemd/aziot-identityd-startup-timeout-lte.conf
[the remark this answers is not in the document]
mlilien: > still don't understand why we need this. as you wrote in the omnect-os PR **aziot-identityd should start after timesync and thus should have network**.

mlilien (commented):
https://github.com/omnect/meta-omnect/pull/300#discussion_r1318143706
https://github.com/omnect/meta-omnect/pull/300#discussion_r1318144895

recipes-core/systemd/systemd/aziot-identityd-startup-timeout-lte.conf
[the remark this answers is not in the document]
HarryWaschkeit: ok, I found the root cause for the problem:

[root@hwt-rpi4-test ~]# systemctl list-dependencies --reverse systemd-networkd-wait-online.service 
systemd-networkd-wait-online.service
* `-network-online.target
*   |-aziot-edged.service
*   |-deviceupdate-agent.service
*   |-docker.service
*   |-iot-client-template-rs.service
*   `-omnect-device-service.service

this means that the identity service is lacking an important dependency, like we already suspected, in particular "Wants=network-online.target" afaics.
due to this lack the service is starting too early and it really depends on the LTE network how fast it connects to the internet whether the service can successfully start or not.
this also explains why we didn't see this problem via Ethernet uplink.

conf/machine/rpi4-test.extra.conf
HarryWaschkeit: then there's no point in using OMNECT_ETH0 and friends any longer afaics, instead the devices can be directly written to OMNECT_ONLINE_INTERFACES in the machine config.

HarryWaschkeit: changed

recipes-core/systemd/systemd/aziot-identityd-startup-timeout-lte.conf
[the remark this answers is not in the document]
HarryWaschkeit: moved condition to meta-omnect

conf/machine/include/rpi.inc
```diff
@@ -18,9 +18,6 @@ CMDLINE_ROOTFS = ""
 # u-boot config vars
 UBOOT_FDT_ADDR = "fdt_addr"
 
-# ethernet device name
-OMNECT_ETH0 = "eth0"
-
 # wlan device name
 OMNECT_WLAN0 = "wlan0"
```
mlilien, line 22: when we remove `OMNECT_ETHX`, why don't we remove `OMNECT_WLAN0` as well?. as far as i understand, systemd will name it `wlan0` anyways.

recipes-connectivity/networkmanager/files/NetworkManager.conf
```diff
@@ -0,0 +1,8 @@
+[main]
+dns=systemd-resolved
+plugins=keyfile
+autoconnect-retries-default=0
```
mlilien, line 4: reading the documentation 0 seems unusual here. what does it mean? no retry?

recipes-connectivity/networkmanager/networkmanager_%.bbappend
```diff
@@ -0,0 +1,18 @@
...
+#       is already there
+
+PACKAGECONFIG:append = " modemmanager"
+PACKAGECONFIG:remove = " dnsmasq"
+
+do_install:append() {
+    install -m 0600 ${WORKDIR}/NetworkManager.conf ${D}${sysconfdir}/NetworkManager/
+}
+
+ALTERNATIVE_PRIORITY[resolv-conf] = "10"
```
mlilien, line 18: does this mean, if i don't use wwan i have no dns resolver?

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -90,18 +88,38 @@ do_install:append:phyboard-polis-imx8mm-4() {
...
+        param_list += '--interface={} '.format(i)
+    return param_list
+
+ONLINE_INTERFACE_ARGS = "${@online_ifc_list_to_parameter_list(d, 'OMNECT_ONLINE_INTERFACES')}"
+
+do_install:append() {
+    sed -i -e 's#^ExecStart=\(.*\)#ExecStart=/bin/bash -c \x27\1 \${OMNECT_WAIT_ONLINE_INTERFACES:-${ONLINE_INTERFACE_ARGS}} --timeout=\${OMNECT_WAIT_ONLINE_TIMEOUT_IN_SECS:-300}\x27#' \
         ${D}${systemd_system_unitdir}/systemd-networkd-wait-online.service
+    if ${@bb.utils.contains('DISTRO_FEATURES', '3g', 'true', 'false', d)}; then
+	install -d ${D}${sysconfdir}/systemd/system/aziot-identityd.service.d
```
mlilien, line 111: we shouldn't use `sysconfdir`  ... that's from a "systemd admin" point of view for the customer. 
we shouldn't install overrides ourselves.

recipes-core/systemd/systemd/aziot-identityd.conf
```diff
@@ -0,0 +1,3 @@
+[Unit]
+Requires=network-online.target
+After=network-online.target
```
mlilien, line 3: this is an override right? ... we shouldn't do that via override.
why don't we set at least the `Requires` for all images?

recipes-connectivity/networkmanager/files/NetworkManager.conf
HarryWaschkeit: according to https://developer-old.gnome.org/NetworkManager/stable/settings-connection.html a value of 0 for autoconnect-retries means "retry forever".
interestingly this cannot be found in the documentation of NetworkManager.conf itself.

recipes-connectivity/networkmanager/networkmanager_%.bbappend
HarryWaschkeit: as long as wwan is not connected Ethernet settings would be used.
so, if eth0 gets dns and gateway settings via dhcp everything works. that is exactly why I needed to deactivate gateway settings in pfsense for rpi4-onect-lab duts.
not sure I understood your question correctly, though.

recipes-core/systemd/systemd/aziot-identityd.conf
HarryWaschkeit: moved both to aziot-idnentityd.inc

recipes-core/systemd/systemd_%.bbappend
HarryWaschkeit: now only aziot-identityd.inc is used.

PR #301, opened 2023-07-03
omnect-os-image: add regression test for packages defined in variables OMNECT_TOOLS and OMNECT_DEVEL_TOOLS

HarryWaschkeit:


recipes-omnect/images/omnect-os-image.bb
```diff
@@ -118,3 +118,72 @@ python () {
...
+        if ! grep -q "^$p " "$manifest"; then
+            bberror "Required TOOL not installed in image: $p"
+            ret=1
+         fi
+    done
+    return $ret
+}
+
+# negative test for packages, i.e. check if _not_ installed
+check_uninstalled_packages() {
```
mlilien, line 139: isn't that the exact function as check_installed_packages without negating the result of grep? so i could negate the result of check_installed_packages?

HarryWaschkeit: nope, both functions fail already for one package not meeting the condition, which is only identical to failing for all packages if the list contains only one packet.

mlilien: :thumbsup:
what if such a packet is already a dependency of a system package?

mlilien (commented):
https://github.com/omnect/meta-omnect/pull/301#discussion_r1251605675

recipes-omnect/images/omnect-os-image.bb
HarryWaschkeit: that is a good question, but currently I don't have another answer than that we will have to talk about the regression test again then ;-)
on the other hand I can't imagine such a dependency that we would want to overrule.
or put in other words: if such a dependency arises it doesn't make sense anymore to include such packets in OMNECT_DEVEL_TOOLS what would fail for release images. and OMNECT_TOOLS isn't critical in this respect for its nature of biegn treated as a positive list only.

mlilien: if bsp A depends on a package and bsp B doesnt ... removing the package from OMNECT_(DEVEL)_TOOLS seems not to be feasible. 
imho we should test that all package are installed we want.
for testing if packages are not installed in case of release vs OMNECT_DEVEL_TOOLS, we imho should only fail, if all of them are installed.

HarryWaschkeit: hmm, so what do we gain from this regression test at all given that one missing tool is already sufficient?
main purpose and root cause for all of that was to verify that the development tools don't get included into release images by accident, with this softened criterion we could just as well do without that test imho.

mlilien: i disagree. @JanZachmann

HarryWaschkeit: if we really want to prepare for this - to me at the moment theoretical - situation right now, instead of soften up the check to the verge of usefulness I would propose a variable defining allowed development tools, something like OMNECT_OS_DEVEL_TOOLS_ALLOWED, which is taken into account during negative checks.
with that approach we would keep control of what will get introduced into release images while retaining a useful test.
just my 0,02 €.

JanZachmann: I understand that the test isn't able to deal with tools that are already deps of other packages and those are even different between BPS's. Or is there a another weak point?
Imho I would prefer to live with the test anyway since:
- I also think the case is very unlikely
- I expect that we will have tools container images soon. thus we'll be able to remove all the tools handling and testing from meta-omnect
- In case I'm wrong regarding both upper points, we would need to refine/refactor the test

JanZachmann:
@HarryWaschkeit I remove myself as reviewer. Just come back to me if necessary.

recipes-omnect/images/omnect-os-image.bb
HarryWaschkeit: rework complete, please review

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
...
+## [kirkstone-0.18.31] Q3 2023
+- kas: removed no longer wanted variables OMNECT_TOOLS & OMNECT_DEVEL_TOOLS from omnect-os.yaml
+- omnect-os-image:
+  - moved packages previously defined in OMNECT_TOOLS_DEFAULT directly into IMAGE_INSTALL
+    and removed variable completely
+    (definition of tools via kas no longer supported, so no fallback needed anymore)
+  - added regression test for packages defined in variable OMNECT_DEVEL_TOOLS which checks
+    for presence/absence of tools depending on image type (release/devel)
+- omnect-os-distro.conf:
+  - renamed variable OMNECT_DEVE_TOOLS_DEFAULT to OMNECT_DEVEL_TOOLS
```
mlilien, line 17: typo

recipes-core/os-release/os-release.bbappend
```diff
@@ -17,7 +17,6 @@ OS_RELEASE_FIELDS += "OMNECT_OS_GIT_BRANCH"
 OS_RELEASE_FIELDS += "OMNECT_OS_GIT_REPO"
 OS_RELEASE_FIELDS += "OMNECT_OS_VERSION"
 
-OS_RELEASE_FIELDS += "OMNECT_TOOLS"
 OS_RELEASE_FIELDS += "OMNECT_DEVEL_TOOLS"
```
mlilien, line 20: imho should only be added in non-release image

recipes-omnect/images/omnect-os-image.bb
```diff
@@ -43,13 +43,12 @@ IMAGE_INSTALL = "\
     polkit \
     u-boot-fw-utils \
     systemd-analyze \
+    e2fsprogs-tune2fs \
+    jq \
+    smartmontools \
```
mlilien, line 48: imho remove it for now ... e.g. we can discuss if we add it for devices with non emmc storage
@JanZachmann?

recipes-omnect/images/omnect-os-image.bb
```diff
@@ -118,3 +117,60 @@ python () {
...
+
+    [ "$release" = 1 ] && relhint="release "
+    bbnote "NOTE: checking for OMNECT_DEVEL_TOOLS in ${relhint}image ..."
+    bbnote "[$tools]"
+
+    set -- ${tools}
+    if [ $# -gt 0 ]; then
+        if [ "$release" = 1 ]; then
+            if check_installed_packages "$manifest" "$tools" 0; then
+                bbwarn 'OMNECT_DEVEL_TOOLS are missing in image!'
```
mlilien, line 158: this is the `release = 1` case ... so that should be the case

recipes-omnect/images/omnect-os-image.bb
JanZachmann: @mlilien you only talk about smartmontools correct? yes, let's remove and maybe bring back in a container someday

recipes-omnect/images/omnect-os-image.bb
```diff
@@ -118,3 +117,60 @@ python () {
 }
 
 inherit omnect_user
+
+inherit logging
+
+# postive test for packages, i.e. check if installed
```
JoergZeidler, line 123: typo: postive -> positive

CHANGELOG.md
HarryWaschkeit: fixed

recipes-core/os-release/os-release.bbappend
HarryWaschkeit: agreed

recipes-omnect/images/omnect-os-image.bb
HarryWaschkeit: uhh, messages were swapped ...

recipes-omnect/images/omnect-os-image.bb
HarryWaschkeit: done

recipes-omnect/images/omnect-os-image.bb
HarryWaschkeit: fixed

mlilien (commented):
``` 
arcel@xmg:~/dev/omnect/git/omnect-os/meta-omnect$ grep -rs OMNECT_DEVEL_TOOLS_DEFAULT .
./CHANGELOG.md:  - renamed variable OMNECT_DEVEL_TOOLS_DEFAULT to OMNECT_DEVEL_TOOLS
./CHANGELOG.md:   named environment variable or tools from OMNECT_DEVEL_TOOLS_DEFAULT as
./.git/logs/HEAD:5734187d9d6636fb715503e31a553428a1bd21d0 d1a4d1aa63da49fbd0c0b8c7c447bc57f5abf781 Marcel Lilienthal <134974+mlilien@users.noreply.github.com> 1671796411 +0100 commit: eg500: add efibootmgr to OMNECT_DEVEL_TOOLS_DEFAULT
./.git/logs/HEAD:d1a4d1aa63da49fbd0c0b8c7c447bc57f5abf781 4f12db469e0c2ae29984b36b8d2a1bc38b0e8191 Marcel Lilienthal <134974+mlilien@users.noreply.github.com> 1671797420 +0100 commit: eg500: add efibootmgr to OMNECT_DEVEL_TOOLS_DEFAULT
./.git/logs/refs/heads/welotec_integration:5734187d9d6636fb715503e31a553428a1bd21d0 d1a4d1aa63da49fbd0c0b8c7c447bc57f5abf781 Marcel Lilienthal <134974+mlilien@users.noreply.github.com> 1671796411 +0100       commit: eg500: add efibootmgr to OMNECT_DEVEL_TOOLS_DEFAULT
./.git/logs/refs/heads/welotec_integration:d1a4d1aa63da49fbd0c0b8c7c447bc57f5abf781 4f12db469e0c2ae29984b36b8d2a1bc38b0e8191 Marcel Lilienthal <134974+mlilien@users.noreply.github.com> 1671797420 +0100       commit: eg500: add efibootmgr to OMNECT_DEVEL_TOOLS_DEFAULT
./conf/tpm-cfg.conf:OMNECT_DEVEL_TOOLS_DEFAULT += "tpm2-tools"
./conf/machine/phyboard-polis-imx8mm-4.extra.conf:# since we do install a lot of dev packages in OMNECT_DEVEL_TOOLS_DEFAULT
```
can you adapt the files grepped in the last 2 lines as well?

mlilien (approved):

JoergZeidler (approved):

PR #309, opened 2023-07-12
added documentation for the welotec eg500 device

mlilien:


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [kirkstone-0.19.3] Q3 2023
+- added documentation for the welotec eg500 device
+- added documentation how to flash on systems which use grub or uboot as last bootloader
```
JoergZeidler, line 10: imho, we should remove the "last" from the bootloader context in the documentation.

HarryWaschkeit: I think Marcel put that there intentionally because more often than not U-boot is not the first bootloader: for raspi there's the binary blob that starts U-Boot and for welotec we have the UEFI firmware, only the tauri directly starts from U-Boot as built by our OS.

JanZachmann:
@mlilien I close this PR in order to fix findings. Findings are fixed with #312

PR #322, opened 2023-08-30
fixed flashmode 2 for devices using efi and grub

mlilien:


recipes-omnect/initrdscripts/omnect-os-initramfs/flash-mode-2
```diff
@@ -122,32 +119,32 @@ run_bmap_tool() {
...
-        root_a_p=${root_a_p::-1}
-        run_cmd sed "s#@@ROOT_DEVICE@@#${root_a_p}#g" /etc/omnect/grub.cfg.in > /tmp/boot/EFI/BOOT/grub.cfg
-        run_cmd umount /tmp/boot
+    local dest_blk=$(readlink -f /dev/omnect/rootA)
+    dest_blk=${dest_blk::-1} # can be /dev/nvme0n1p or /dev/sda
+    local p=${dest_blk:0-1}
+    if [ "${p}" == "p" ]; then
+        dest_blk=${dest_blk::-1}
+    else
+        p=""
```
JoergZeidler, line 143: just to clarify: only needed for efi_grub handling?

mlilien: yes its only needed for efi+grub handling.

but imho can not be moved. in flash-mode-1 these params are already known.
this was my idea to reuse as much code as i can regarding the efi+grub handling in both flash-modes.

perhaps a better way would be to have dedicated flash-mode implementations for u-boot vs efi+grub devices, but this imho worth a dedicated pbi. if we use dedicated implementations, we should do this for all scripts.
since the handling of reading/writing grub env vars is complicated, we would have to split most script files into multiple files (common vs special parts).  currently this is not in focus for this pbi.

JoergZeidler: ok, let's use it at the moment as it is.

JoergZeidler (approved):

PR #328, opened 2023-09-15
eg500 bootloader update

mlilien:


recipes-omnect/images/omnect-os-update-image.bb
```diff
@@ -11,15 +11,21 @@ LIC_FILES_CHKSUM = "\
 
 SRC_URI = "file://sw-description"
 
-COMPATIBLE_MACHINE = "rpi|phytec-imx8mm"
+COMPATIBLE_MACHINE = "rpi|phytec-imx8mm|omnect_grub"
```
JoergZeidler, line 14: Think we already talked about the compatible machine mechanism. Should we now use omnect_uboot instead of rpi and phytec?

mlilien: we can not. i changed the override order in this PR that "rpi" and "phytec-imx8mm" will supersede "omnect_uboot".
rpi and phytec machine configs both set "omnect_uboot" and we use it in other occurrences. 
Using "omnect_uboot" here you could not distinguish between rpi and phytec. We could use "omnect_uboot" for the phytec devices, but it feels wrong. Imho we would need something like "omnect_uboot_imx" then.

mlilien: Nevermind you're right. For the variable `COMPATIBLE_MACHINE` we could do that. For the install/deploy steps and the sw-description not. I would prefer if we use it analog to the sw-description here.

JoergZeidler: yes, that sounds reasonable.

recipes-omnect/images/omnect-os-update-image.bb
```diff
@@ -36,18 +42,27 @@ do_bootloader_package:rpi() {
     tar -czvf boot-partition-update.tar.gz -C ${DEPLOY_DIR_IMAGE}/boot-partition .
 }
 
-do_bootloader_package() {
+do_bootloader_package:omnect_uboot() {
```
JoergZeidler, line 45: imho we should  here use :phytec-imx8mm

HarryWaschkeit: nope, that would not work for raspberry devices, omnect_uboot is the right one imo.

mlilien: done

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_v1_grub.sh
```diff
@@ -94,13 +88,16 @@ fi
 
 if [[ $action == "apply" ]]; then
     echo "Applying update." >> "${log_dir}/swupdate.log"
-    mount_boot && grub-editenv /boot/EFI/BOOT/grubenv set omnect_validate_update_part=${update_part} && umount /boot
+    if [ ! -f "/run/omnect-bootloader-update" ]; then
```
JoergZeidler, line 91: like the new temporary file name. would you like to change the u-boot file name (/run/bootloader/update) as well?

recipes-omnect/images/omnect-os/omnect_grub/sw-description
```diff
@@ -0,0 +1,69 @@
...
+                fh:close()
+                swupdate.trace(\"bootloader_version: \" .. bootloader_version)
+                swupdate.trace(\"image.version: \" .. image.version)
+                if bootloader_version == image.version then
+                result = nil
+                else
+                f = io.open(\"/run/omnect-bootloader-update\", \"w\")
+                f:close()
+                result = image
+                end
```
HarryWaschkeit, line 64: I find missing indentation irritating here; is there any reason to refrain from indenting?

JoergZeidler (approved):

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_v1_grub.sh
mlilien: done

recipes-omnect/images/omnect-os/omnect_grub/sw-description
mlilien: fixed

HarryWaschkeit (approved):

PR #331, opened 2023-09-25
welotec eg500: refactored grub.cfg used in wic image

mlilien:


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [kirkstone-0.22.2] Q3 2023
+- welotec eg500: refactored wic image to use the correct grub.cfg;
+  NOTE: you have to inject `recipes-omnect/grub-cfg/grub-cfg/grub-usb.cfg`
+  into the `boot` partition `/EFI/BOOT/grub.cfg` now, if you want to be able
+  to boot from usb.
+  (this enables flash-mode-2 to flash the correct grub.cfg independent of the
+  used omnect-os-initramfs)
```
HarryWaschkeit, line 14: if I understand that correctly: this is something we definitely should also put in a user documentation or otherwise a user would not easily figure why the built image cannot be used on a usb stick out-of-the-box.
just as a side note ...

HarryWaschkeit (approved):

CHANGELOG.md
JanZachmann: Imho there are multiple issues with documentation:

- in README.md we have a chapter "Runtime configuration" describing iot-identity-service and azure-device-update config injection. Now we have another mandatory run time config for a specific board that shouldn't be documeted in Readme.md
- README.device.md is imho not the place we should document ALL device specific stuff anymore. It might be still good for a rough overview and for comparison  of high level attributes and features. Imho we should go for a new structure with a dedicated subpage for each hardware platform. I created a [PBI](https://dev.azure.com/conplementag/Omnect/_backlogs/backlog/Omnect%20Team/Backlog%20items/?workitem=36517)
- omnect-cli docu must also be updated since it contains u-boot specific examples.

JanZachmann (approved):

PR #352, opened 2023-10-16
Swupdate handler v2

JoergZeidler:
tested with omnect-os PR https://github.com/omnect/omnect-os/pull/354 on

rpi4
eg500
tauri


recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -0,0 +1,757 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
```
mlilien, line 4: add our copyright information as well?

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -0,0 +1,757 @@
...
+log_error_pref="\033[1;31m[E]\033[0m"
+
+#
+# Files and Folders information
+#
+workfolder=
+output_file=/var/log/aduc-logs/swupdatehandlerv2.output
+log_file=/var/log/aduc-logs/swupdatehandlerv2.log
+swupdate_log_file=/var/log/aduc-logs/swupdate.log
+result_file=/var/log/aduc-logs/swupdatehandlerv2.result
```
mlilien, line 50: can we add timestamp suffixes ... so i don't overwrite logs when doing multiple runs of install, e.g. without apply?

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -0,0 +1,757 @@
...
+fi
+
+if [ -n "$do_install_action" ]; then
+    InstallUpdate
+    exit $ret_val
+fi
+
+if [ -n "$do_apply_action" ]; then
+    ApplyUpdate
+    exit $ret_val
```
mlilien, line 749: i can theoretically use both parameter in one step ... `--action-install` and `--action-apply` but only the install would be applied and the script would return with the result code of the install step. 
Same finding for other usage of multiple other parameters as well.

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update_git.bb
```diff
@@ -116,14 +112,6 @@ do_install:append() {
   install -m 0770 -o adu -g adu /dev/null ${D}${sysconfdir}/omnect/consent/swupdate/installed_criteria
 }
 
-do_install:append:omnect_grub() {
-  install -m 0755 ${WORKDIR}/swupdate_v1_grub.sh ${D}${libdir}/adu/adu-swupdate.sh
-}
```
mlilien, line 121: please install the handler v2 at least for the developer image ... e.g. that i can use .swu files directly.
@JanZachmann ?

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
JoergZeidler: imho, the log files won't be overwritten. The script will always append the log info e.g. "echo "xxx" **>>** "

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
JoergZeidler: imho, the script will only be used in the automated environment like swupdate_handler_v2 or in the smoke-tests...

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update_git.bb
JanZachmann: my proposal: only for devel imgs install `swupdate-handlerv2.sh` to /usr/lib/`swupdate.sh` 

pros:
- we need to maintain only a single file (instead of two if we would have v1 and v2 script)
- we can reference a dedicated script in [omnect-os readme](https://github.com/omnect/omnect-os/pull/354/files#r1363715975) (but for devel imgs only)

cons:
- ?

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
mlilien: does upstream use multiple log files, or did we introduce that or added to that? `log_file` vs `swupdate_log_file` (vs. `output_file`)?

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
JoergZeidler: yes, i saw now that our old v1 script had also the MS and the Conplement  copyright info.

@JanZachmann FYI.

JoergZeidler: done

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update_git.bb
JoergZeidler: done

JanZachmann (approved):

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update_git.bb
mlilien: can we move that /usr/lib/swupdate/swupdate.sh ... (the dir already exists)?

JoergZeidler: yes we can move, but we need an decision. 

should it available under the new path:
**/usr/lib/swupdate/**
or use the old path
**/usr/lib/adu/**


In case **/usr/lib/swupdate/**, we need a new file name, swupdate.sh already exists  under that path.

@JanZachmann @mlilien

JanZachmann: @JoergZeidler @mlilien 

`/usr/lib/swupdate/omnect-swupdate.sh`?

JoergZeidler: done

mlilien (approved):

PR #363, opened 2023-11-07
SBOM handling + reviewed and removed dependencies

mlilien:


classes/dependency-track.bbclass
```diff
@@ -0,0 +1,88 @@
...
+            purl = 'pkg:yocto/{}@{}'.format(cpe_split[4],version)
+        else:
+            purl = 'pkg:yocto/{}/{}@{}'.format(cpe_split[3],cpe_split[4],version)
+        if not next((c for c in sbom["components"] if c["cpe"] == cpe), None):
+            sbom["components"].append({
+                "name": names[index],
+                "version": version,
+                "cpe": cpe,
+                "purl": purl
+            })
```
HarryWaschkeit, line 67: better throw error if a duplicate entry is detected instead of silently discarding any additional entries?

mlilien: if i do that, i get errors for the following packages:
``` 
ERROR: libtool-2.4.7-r0 do_dependencytrack_collect: cpe handled twice: cpe:2.3:a:*:libtool:2.4.7:*:*:*:*:*:*:*
ERROR: libgcc-11.4.0-r0 do_dependencytrack_collect: cpe handled twice: cpe:2.3:a:*:gcc:11.4.0:*:*:*:*:*:*:*
ERROR: libgcc-initial-11.4.0-r0 do_dependencytrack_collect: cpe handled twice: cpe:2.3:a:*:gcc:11.4.0:*:*:*:*:*:*:*
ERROR: gcc-source-11.4.0-11.4.0-r0 do_dependencytrack_collect: cpe handled twice: cpe:2.3:a:*:gcc:11.4.0:*:*:*:*:*:*:*
```
it's not clear yet why.

mlilien: a recipe with one `CPE_PRODUCT` setting can result in multiple packages

JanZachmann (approved):

classes/dependency-track.bbclass
HarryWaschkeit: if run time mattered the loop could be optimized a bit by checking existence in sbom before calculating sbom fields, but I doubt that it saves significantly time in comparison to the linear search in the increasing list.

HarryWaschkeit (approved):

PR #371, opened 2023-12-01
rust 1.74.0

mlilien:


kas/distro/oe.yaml
```diff
@@ -20,6 +20,6 @@ repos:
     patches:
       p001:
         repo: "meta-omnect"
-        path: "kas/patches/oe_bootimg_efi_wic_and_image_bbclass.patch"
+        path: "kas/patches/oe.patch"
```
mlilien, line 23: FYI: @HarryWaschkeit I refactored the patch here. oe.patch is concatenated of the other oe_*.patch files. 
i think it's easier that way, if you have doubts, please comment.

JanZachmann (approved):

PR #377, opened 2023-12-13
cargo: disabled creating path overrides

mlilien:


classes/omnect_rust_azure-iot-sdk_deps.bbclass
```diff
@@ -23,7 +23,7 @@ eis-utils = { path = "${WORKDIR}/eis-utils" }
 [patch.'https://github.com/omnect/azure-iot-sdk-sys']
 azure-iot-sdk-sys = { path = "${WORKDIR}/azure-iot-sdk-sys" }
 
-[patch.'https://git@github.com/Azure/iot-identity-service.git']
+[patch.'https://github.com/Azure/iot-identity-service.git']
```
mlilien, line 26: @empwilli with this PR your warnings should disappear. this bbclass file is now not shared anymore. before it was used by `omnect-device-service` and `iot-client-template-rs`. we decided to leave it as bbclass anyway, if we reintroduce `iot-client-template-rs`.

JanZachmann (approved):

PR #385, opened 2024-01-24
omnect-device-service + wifi-commissiong-gatt-service: deploy PV

mlilien:


recipes-omnect/omnect-device-service/omnect-device-service.inc
```diff
@@ -67,6 +67,14 @@ do_install:append:omnect_uboot() {
     install -m 0640 -D ${S}/sudo/omnect-device-service-uboot ${D}${sysconfdir}/sudoers.d/omnect-device-service-bootloader-env
 }
 
+# for post process handling we deploy PV
```
HarryWaschkeit, line 70: out of curiosity: where does this post processing happen?
maybe it is worth mentioning here, too?

recipes-omnect/omnect-device-service/omnect-device-service.inc
```diff
@@ -67,6 +67,14 @@ do_install:append:omnect_uboot() {
...
+inherit deploy
+do_deploy() {
+    echo "${PV}" > "${DEPLOYDIR}/omnect-device-service.version"
+}
+addtask do_deploy after do_compile before do_build
+
+
```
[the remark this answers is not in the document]
mlilien: done

recipes-omnect/omnect-device-service/omnect-device-service.inc
mlilien: not in meta-omnect. it is for post processing which doesn't affect the content of the image.
should i remove or rephrase the comment? if rephrase, to what?

HarryWaschkeit: for an omnect developer like me who doesn't know what that post processing exactly is, a comment would be helpful indeed, but as it's a public repository all other audience doesn't need to (and maybe wouldn't) understand what is meant by it.
so, it's ok for me to leave it as is.

HarryWaschkeit (approved):

mlilien:
@HarryWaschkeit I thought about your comment and will solve it another way.

PR #398, opened 2024-02-27
feat(iot-hub-device-update): only allow forward updates

JoergZeidler:
With this change we enforce forward updates on device side. In all other cases the update will fail. Additionally the implementation of another version check in azure cloud is planned, so that downgrades cannot be deployed to devices at all.


JanZachmann (changes requested):
PR title:
feat: iot-hub-device-update: only allow forward updates on the device ->
feat(iot-hub-device-update): only allow forward updates

PR desc:
With this change we enforce forward updates on device side. In all other cases the update will fail. Additionally the implementation of another version check in azure cloud is planned, so that downgrades cannot be deployed to devices at all.

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -513,6 +602,39 @@ CheckIsInstalledState() {
     result "$aduc_result"
 }
 
+#
+# Implementation of 'IsInstallable' function, for the reference Yocto Image.
```
mlilien, line 606: what reference Yocto Image? ... this is intended for an omnect-os image (same finding for other script)

JoergZeidler: done

mlilien (approved):

JanZachmann (approved):

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -479,13 +485,112 @@ function is_installed() {
...
+        v1=("${BASH_REMATCH[@]}")
+        if [[ "$2" =~ $regex ]]
+        then
+            v2=("${BASH_REMATCH[@]}")
+        else
+            return 0    # version format illegal
+        fi
+    else
+        return 0        # version format illegal
+    fi
```
JanZachmann, line 520: better check v1 and then v2 on the same level. no need to check v2 in if of v1 check

JanZachmann: since you don't echo an error you could also have only one if where checking both, v1 and v2

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -479,13 +485,112 @@ function is_installed() {
...
+# If version1 older than version2           -> return 2
+# If version1 equal version2                -> return 3
+# Illegal format in version1 or version2    -> return 0
+#
+# Usage: version_comp $version1 $version2
+#
+# shellcheck disable=SC2034
+function version_comp () {
+
+    local regex='^([0-9]+)\.([0-9]+)\.([0-9]+)\.([0-9]+)$'
```
JanZachmann, line 505: why don't we exclude leading zeros here?

JanZachmann (changes requested):

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
JoergZeidler: changed as discussed

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
JoergZeidler: removed sed and changed regex

HarryWaschkeit (commented):
imho it's a bit of a pity to have two versions of the update script which are nearly identical but I'm not sure whether it's worth reworking that; maybe if the script becomes even more complex in the future.

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -479,13 +485,112 @@ function is_installed() {
...
+#
+# Usage: is_installable $installedCriteria $software_version_file <out resultCode> <out extendedResultCode> <out resultDetails>
+#
+# shellcheck disable=SC2034
+function is_installable() {
+    local -n rc=$3  # name reference for resultCode
+    local -n erc=$4 # name reference for extendedResultCode
+    local -n rd=$5  # name reference for resultDetails
+    local current_version=$(cat $software_version_file | head -n1 | awk '{print $2;}')
+    local new_version=$(echo $1 | head -n1 | awk '{print $2;}')
```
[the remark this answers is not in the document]
JoergZeidler: done

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -513,6 +618,39 @@ CheckIsInstalledState() {
...
+#
+CheckIsInstallableState() {
+    log_info "IsInstallableTask(\"$1\"), adu-version path:\"$software_version_file\""
+
+    local local_rc=2
+    local local_erc=3
+    local local_rd="4"
+    local aduc_result=""
+
+    # Evaluates installedCriteria string.
```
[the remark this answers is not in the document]
HarryWaschkeit: ok, it's implemented that way in several places, also /wo comment, so it doesn't help a lot to improve it here either.

HarryWaschkeit (approved):

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -479,13 +485,108 @@ function is_installed() {
...
+# If version1 older than version2           -> return 2
+# If version1 equal version2                -> return 3
+# Illegal format in version1 or version2    -> return 0
+#
+# Usage: version_comp $version1 $version2
+#
+# shellcheck disable=SC2034
+function version_comp () {
+
+    local regex='^0*([0-9]+)\.0*([0-9]+)\.0*([0-9]+)\.0*([0-9]+)$'
```
JanZachmann, line 505: so we allow leading zeros but ignore it?
can you please add some examples for valid versions?

JanZachmann (changes requested):

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
JoergZeidler: https://regex101.com/r/rhe1hM/1

JanZachmann (dismissed):

mlilien (dismissed):

HarryWaschkeit (approved):

PR #401, opened 2024-02-29
fix(u-boot): don't use redundant-env-fragment.cfg for bootloader checksum computation

mlilien:
redundant-env-fragment.cfg is generated during build from redundant-env-fragment.cfg.template

### DIFF
--- a/recipes-bsp/u-boot/u-boot_%.bbappend
+++ b/recipes-bsp/u-boot/u-boot_%.bbappend
@@ -37,6 +37,9 @@ OMNECT_BOOTLOADER_CHECKSUM_FILES += "${OMNECT_THISDIR_SAVED}/u-boot-scr.bb"
 OMNECT_BOOTLOADER_CHECKSUM_FILES += "${OMNECT_THISDIR_SAVED}/u-boot/*"
 
 OMNECT_BOOTLOADER_CHECKSUM_FILES_GLOB_IGNORE = "${OMNECT_THISDIR_SAVED}/u-boot/.gitignore"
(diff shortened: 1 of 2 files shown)


recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -37,6 +37,9 @@ OMNECT_BOOTLOADER_CHECKSUM_FILES += "${OMNECT_THISDIR_SAVED}/u-boot-scr.bb"
 OMNECT_BOOTLOADER_CHECKSUM_FILES += "${OMNECT_THISDIR_SAVED}/u-boot/*"
 
 OMNECT_BOOTLOADER_CHECKSUM_FILES_GLOB_IGNORE = "${OMNECT_THISDIR_SAVED}/u-boot/.gitignore"
+# don't include files which are generated on build:
+OMNECT_BOOTLOADER_CHECKSUM_FILES_GLOB_IGNORE += "${OMNECT_THISDIR_SAVED}/u-boot/redundant-env-fragment.cfg"
```
mlilien, line 41: other option would be, we refactor so that redundant-env-fragment.cfg is generated in `WORKDIR`

HarryWaschkeit (approved):

recipes-bsp/u-boot/u-boot_%.bbappend
HarryWaschkeit: either way is fine for me; if we faced lots of such exceptions in future maybe having that stuff in WORKDIR would be preferable

PR #440, opened 2024-06-05
fix: os-release entry in target architecture

empwilli:
The omnect-os-distro.conf recipe is evaluated with `TARGET_ARCH` set to `allarch` (as the recipe is architecture independent). We instead introduce a new variable `OMNECT_TARGET_ARCH` that is populated earlier with the actual target architecture and export this variable instead of `TARGET_ARCH` in os-release.


conf/distro/include/omnect-os-distro.conf
```diff
@@ -160,3 +160,7 @@ OMNECT_DEVEL_TOOLS = "\
 "
 
 OMNECT_DEVEL_TOOLS:append:x86-64 = "acpica dmidecode"
+
+# make target architecture for visible for later processing
```
mlilien, line 164: -> "make target architecture ~~for~~ visible for later processing, e.g. in os-release"

empwilli: ack

JoergZeidler (approved):

PR #450, opened 2024-06-17
feat: enabled disabling bootloader version check

mlilien:
feat: enabled disabling bootloader version check by setting `OMNECT_BOOTLOADER_VERSION_CHECK_DISABLE=1`

### DIFF
--- a/classes/omnect_bootloader_versioning.bbclass
+++ b/classes/omnect_bootloader_versioning.bbclass
@@ -93,7 +93,6 @@ python() {
             bb.debug(1, "overriding version_checksum %s -> %s" % (version_checksum,version_checksum_override_list[1]))
             version_checksum = version_checksum_override_list[1]
 
-
     version_checksum_expected = d.getVar("OMNECT_BOOTLOADER_CHECKSUM_EXPECTED")
     if not version_checksum_expected:
         bb.fatal("OMNECT_BOOTLOADER_CHECKSUM_EXPECTED not set; computed checksum is: \"%s\"" % version_checksum)
--- a/classes/omnect_bootloader_versioning.bbclass
+++ b/classes/omnect_bootloader_versioning.bbclass
@@ -94,12 +94,14 @@ python() {
             version_checksum = version_checksum_override_list[1]
 
 
-    if "1" != d.getVar("OMNECT_BOOTLOADER_VERSION_CHECK_DISABLE"):
-        version_checksum_expected = d.getVar("OMNECT_BOOTLOADER_CHECKSUM_EXPECTED")
-        if not version_checksum_expected:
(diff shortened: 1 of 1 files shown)


classes/omnect_bootloader_versioning.bbclass
```diff
@@ -93,11 +93,13 @@ python() {
...
-        bb.fatal("OMNECT_BOOTLOADER_CHECKSUM_EXPECTED not set; computed checksum is: \"%s\"" % version_checksum)
-    if version_checksum_expected != version_checksum:
-        bb.fatal("expected bootloader checksum (OMNECT_BOOTLOADER_CHECKSUM_EXPECTED): \"%s\" is different from computed: \"%s\"" % (version_checksum_expected, version_checksum))
+
+    if "1" != d.getVar("OMNECT_BOOTLOADER_VERSION_CHECK_DISABLE"):
+        version_checksum_expected = d.getVar("OMNECT_BOOTLOADER_CHECKSUM_EXPECTED")
+        if not version_checksum_expected:
+            bb.fatal("OMNECT_BOOTLOADER_CHECKSUM_EXPECTED not set; computed checksum is: \"%s\"" % version_checksum)
+        if version_checksum_expected != version_checksum:
+            bb.fatal("expected bootloader checksum (OMNECT_BOOTLOADER_CHECKSUM_EXPECTED): \"%s\" is different from computed: \"%s\"" % (version_checksum_expected, version_checksum))
```
HarryWaschkeit, line 102: imo it would be handy to get the checksum calculated and shown even if version check is disabled.
so it is easy to copy the correct checksum over once all bootloader changes are in place instead of needing another failing build round.

```suggestion
    version_checksum_expected = d.getVar("OMNECT_BOOTLOADER_CHECKSUM_EXPECTED")
    if not version_checksum_expected:
        bb.fatal("OMNECT_BOOTLOADER_CHECKSUM_EXPECTED not set; computed checksum is: \"%s\"" % version_checksum)
    if version_checksum_expected != version_checksum:
        if "1" != d.getVar("OMNECT_BOOTLOADER_VERSION_CHECK_DISABLE"):
            bb.fatal("expected bootloader checksum (OMNECT_BOOTLOADER_CHECKSUM_EXPECTED): \"%s\" is different from computed: \"%s\"" % (version_checksum_expected, version_checksum))
        else:
            bb.error("expected bootloader checksum (OMNECT_BOOTLOADER_CHECKSUM_EXPECTED): \"%s\" is different from computed: \"%s\"" % (version_checksum_expected, version_checksum))
```

I know in the above suggestion the case that there's no checksum set yet isn't handled quite cleanly but an arbitrary string is easily set so I wouldn't mind

HarryWaschkeit (approved):

PR #459, opened 2024-07-01
chore: bump omnect-device-service and wifi-commissioning-gatt-service

JanZachmann:


recipes-omnect/wifi-commissioning-gatt-service/wifi-commissioning-gatt-service_0.4.2.bb
JanZachmann, line 1: @mlilien I've no idea why this is not displayed as diff

PR #479, opened 2024-07-30
fix(bootloader-versioned): fix attempt for wrong artefact generation

HarryWaschkeit:
Added inheritance of class omnect_bootloader_versioning to recipe to hopefully yield complete dependencies for proper rebuild of versioned bootloader artefact also for 3rd party machines contained is extra meta layers.

### DIFF
--- a/recipes-bsp/u-boot/bootloader-versioned.bb
+++ b/recipes-bsp/u-boot/bootloader-versioned.bb
@@ -16,7 +16,11 @@ OMNECT_BOOTLOADER_EMBEDDED_VERSION_IMAGESIZE ?= ""
 OMNECT_BOOTLOADER_EMBEDDED_VERSION_LOCATION ?= ""
 OMNECT_BOOTLOADER_EMBEDDED_VERSION_IMAGESIZE ?= ""
 
-inherit deploy omnect_uboot_embedded_version
+# although not needed here for version calculation, nevertheless inherit class
+# omnect_bootloader_versioning to (hopefully) cause correct re-building of
+# verisioned bootlaoder artefact instead of matching an old version in SSTATE
+# cache
+inherit deploy omnect_bootloader_versioning omnect_uboot_embedded_version
 
 do_compile:append() {
     cp "${WORKDIR}/omnect_get_bootloader_version.sh.template" "${WORKDIR}/omnect_get_bootloader_version.sh"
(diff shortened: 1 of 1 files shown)


recipes-bsp/u-boot/bootloader-versioned.bb
```diff
@@ -16,7 +16,11 @@ OMNECT_BOOTLOADER_EMBEDDED_VERSION_IMAGESIZE ?= ""
 OMNECT_BOOTLOADER_EMBEDDED_VERSION_LOCATION ?= ""
 OMNECT_BOOTLOADER_EMBEDDED_VERSION_IMAGESIZE ?= ""
 
-inherit deploy omnect_uboot_embedded_version
+# although not needed here for version calculation, nevertheless inherit class
+# omnect_bootloader_versioning to (hopefully) cause correct re-building of
+# verisioned bootlaoder artefact instead of matching an old version in SSTATE
+# cache
+inherit deploy omnect_bootloader_versioning omnect_uboot_embedded_version
```
mlilien, line 23: this will only work in kirkstone because of:
https://github.com/omnect/meta-omnect/blob/main/classes/omnect_bootloader_versioning.bbclass#L30

in scarthgap i removed, that because we have a race condition there.

other possible method is to prevent this recipe from sstate:
```
SSTATE_SKIP_CREATION = "1"
```

mlilien: imho we should add `virtual/bootloader` to `DEPENDS` that should fix the issue.

HarryWaschkeit: u-boot-karo is already part of DEPENDS in line 10 via variable OMNECT_BOOTLOADER_EMBEDDED_VERSION_BBTARGET and it is apparently not sufficient.

mlilien: imho the sstate cache is invalid, when a dependency has an invalid sstate cache.
if sw-update has the updated bootloader artefact, it should be the same as used in the wic-image.
i think we look into the wrong place.

HarryWaschkeit: I tried again with SKIP_SSTATE_CREATION and it yielded the same changes in tasks dependencies now as inheriting omnect_bootloader_versioning did, so I would give that a go.

mlilien (approved):

PR #488, opened 2024-08-09
refactor(kernel): added more kernel options to fragments

HarryWaschkeit:
For an i.MX6 device several kernel options were missing for virtualization (docker) to work.
It seems that kernel configurations for officially supported devices already set the needed configuration options.


recipes-kernel/linux/files/netfilter.cfg
```diff
@@ -205,6 +213,7 @@ CONFIG_NFT_REJECT_IPV4=m
 CONFIG_NFT_DUP_IPV4=m
 CONFIG_NFT_FIB_IPV4=m
 CONFIG_NF_TABLES_ARP=y
+# CONFIG_NF_TABLES_BRIDGE is not set
```
mlilien, line 216: disabled on purpose?

HarryWaschkeit: I double checked that and found that I misinterpreted what make menuconfig tells me about config options.
so, this option is now active and a few more were added, pls review again.

mlilien (approved):

PR #489, opened 2024-08-09
fix(factory-reset): ensure boot partition is unmounted before optional restore handling

mlilien:


recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
```diff
@@ -112,14 +112,17 @@ factory_reset_finish() {
     # set return status
     set_bootloader_env_var ${FACTORY_RESET_STATUS_VAR} "${FACTORY_RESET_RET1}:${FACTORY_RESET_RET2}"
 
+    # we already mounted /dev/omnect/boot when we're using grub. so we unmount.
+    if [ -f /usr/bin/grub-editenv ]; then
```
HarryWaschkeit, line 116: this seems dirty to me, I would consider using existence of files in packages for decision upon one or another bootloader somewhat fragile by itself.
moreover, actually it isn't about bootloader here at all, having mounted /boot or not currently definitely depends on the bootloader, but what we actually do want to ensure is, that /boot gets cleanly unmounted if currently mounted.

as simpler solution, we could use 'mountpoint -q /boot && {sync; run_cmd umount ${ROOTFS_DIR}/boot }'.
(or also w/o 'sync' which should always be implicitly ensured by the kernel before actually unmounting the filesystem.

mlilien: maybe we should refactor this ... i'll not do it for this file only.
we use this mechanism in multiple occurrences.

> moreover, actually it isn't about bootloader here at all, having mounted /boot or not currently definitely depends on the bootloader, but what we actually do want to ensure is, that /boot gets cleanly unmounted if currently mounted.

i disagree. normally /boot shouldn't be mounted at all before fsmount.

@JanZachmann leave this PR as is and create a refactor task for grub uboot env handling?

JanZachmann: > @JanZachmann leave this PR as is and create a refactor task for grub uboot env handling?
ok for me. @HarryWaschkeit do you agree?

JanZachmann (approved):

PR #499, opened 2024-09-19
feat(bootloader-versioned): added support of UUU tag for i.MX production process

HarryWaschkeit:
Boot ROMs of some i.MX processors provide a serial downloader which allows for loading a bootloader over a USB OTG interface, which gets started and takes over the serial communication.
Several protocols are used over that link which allow flashing of images to various bootmedia among other things.
The tool used on a host PC was renamed from originally mfgtool to uuu.
Its sources can be found on https://github.com/nxp-imx/mfgtools, releases can be obtained from https://github.com/nxp-imx/mfgtools/releases.
The latter also provides a PDF for every version which describes this tool in detail.
One requirement for that tool is that the bootloader binary bears some end mark which allow for size calculation of the bootloader.
This change includes support to add such an end mark to the bootloader binary during the course of embedding the (omnect-os) version of it.


classes/omnect_uboot_embedded_version.bbclass
```diff
@@ -58,6 +58,10 @@
 #    up to defined size of resulting file; default value is 0x00
 #    however, for NOR flashes it would be better to have an 0xff, therefore
 #    this variable
+#  - OMNECT_BOOTLOADER_EMBEDDED_VERSION_UUUTAG
+#    if bootloader is used for production programming, i.e. uuu tool flashes
+#    boot medium via ROM serial boot, it needs an end tag for the tool to
+#    detect its size.
```
mlilien, line 64: rest of this PR makes me believe this is a boolean? if so please document, otherwise an example would be nice to understand a possible value.

classes/omnect_uboot_embedded_version.bbclass
```diff
@@ -146,8 +150,22 @@ python omnect_uboot_embed_version() {
         bb.fatal("Unable to stat bootloader binary \"%s\"" % (version_file))
     filesize = st.st_size
 
+    # do we need an end tag which will be searched for by uuu tool during
+    # initial flashing in production process?
+    # NOTE:
+    #   this handling is normally found in yocto layer meta-freescale in
+    #   classes file uuu_bootloader_tag.bbclass.
+    #   in order to be independend of thos meta layer this is handled here,
+    #   too, we're processing the binary here anyway.
```
mlilien, line 159: do we have cases supporting an imx device without meta-freescale?

classes/omnect_uboot_embedded_version.bbclass
```diff
@@ -146,8 +150,22 @@ python omnect_uboot_embed_version() {
         bb.fatal("Unable to stat bootloader binary \"%s\"" % (version_file))
     filesize = st.st_size
 
+    # do we need an end tag which will be searched for by uuu tool during
+    # initial flashing in production process?
+    # NOTE:
+    #   this handling is normally found in yocto layer meta-freescale in
+    #   classes file uuu_bootloader_tag.bbclass.
+    #   in order to be independend of thos meta layer this is handled here,
```
mlilien, line 158: typo "thos"

HarryWaschkeit: fixed.

classes/omnect_uboot_embedded_version.bbclass
HarryWaschkeit: yes, dehndetect does not currently use/need meta-freescale.

classes/omnect_uboot_embedded_version.bbclass
HarryWaschkeit: improved description.

mlilien (approved):

PR #503, opened 2024-10-02
feat(factory-reset): interface change to allow parameterization of files to preserve

mlilien:
- adapt to new interface of factory-reset
- bugfix for u-boot `omnect_update_flow` print if bootloader gets updated
- escaping bugfix for `bootloader_env_*.sh`


conf/machine/include/rpi.inc
```diff
@@ -40,5 +40,4 @@ OMNECT_BOOTLOADER="u-boot"
 # OMNECT_BOOTLOADER_CHECKSUM_EXPTECTED:pn-u-boot - build will fail, if the
 # computed checksum is different to this; set to <oldchecksum> when
 # OMNECT_BOOTLOADER_CHECKSUM_COMPATIBLE:pn-u-boot is set
-OMNECT_BOOTLOADER_CHECKSUM_EXPECTED:pn-u-boot = "7320f8af2c8e3769de2af4a8151b1f06b297beba134db93b0b3c2df9b50b9bcf"
-OMNECT_BOOTLOADER_CHECKSUM_COMPATIBLE:pn-u-boot = "78ea9242f234de05655db0a069bd21a38833611cae3b5efc4f823c19e449cf59 7320f8af2c8e3769de2af4a8151b1f06b297beba134db93b0b3c2df9b50b9bcf"
+OMNECT_BOOTLOADER_CHECKSUM_EXPECTED:pn-u-boot = "f40663d044e6bf3d85adc3f8a082c3fd24523b6befa416c2040d077c896500fd"
```
mlilien, line 43: TODO

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
```diff
@@ -152,41 +122,70 @@ factory_reset_backup() {
...
+    local    do_backup="${1}"
+    local restore_list="${@:2}"
+
     local ret=0
     local status=0
+    local context=""
 
-    restore_list=${restore_list//;/ }
-    for i in ${restore_list}; do
+    for i in ${restore_list[@]}; do
```
HarryWaschkeit, line 166: afaics `restore_list` is no array here, so `[@]` can be omitted, the same applies below.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
```diff
@@ -207,6 +206,34 @@ factory_reset_umount() {
     run_cmd mount
 }
 
+handle_status() {
+    local status=${1}
+    local error=${2}
+    local context=${3}
+    local preserve_paths=${@:4}
+
+    msg "status: status=${status}, error=${error}, context=${context}, preserve_path=${preserve_paths[@]}"
```
HarryWaschkeit, line 215: again, `preserve_paths` is no array, so there's no use of `[@]`

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
```diff
@@ -216,70 +243,78 @@ factory_reset_run() {
...
+    msg "Running factory-reset [mode=${mode}, preserve=${preserve[@]}]..."
+    factory_reset_mount
+
+    # create preserve_list
+    declare -a preserve_list
+    preserve_list+=("/etc/omnect/factory-reset.d/")
+    if [[ " ${preserve[@]} " =~ " applications " ]]; then
+        preserve=( ${preserve[@]/applications} )
+        for i in $(find ${ROOTFS_DIR}/etc/omnect/factory-reset.d/ -iname *.json); do
+            preserve_list+=( $(cat ${i} | jq -r -e '.paths[]') )
```
HarryWaschkeit, line 264: no need for `cat`, could be replaced with simple stdin redirection from file

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
```diff
@@ -216,70 +243,78 @@ factory_reset_run() {
...
+        done
     fi
 
-    msg "Running Factory Reset [mode=${factory_reset_mode}]..."
-
-    if [ -n "${factory_reset_restore_list}" ]; then
-        msg "restore list present"
-        factory_reset_mount  # logical filesystem tree
+    for key in ${preserve[@]}; do
+        preserve_list+=( $(cat ${ROOTFS_DIR}/etc/omnect/factory-reset.json | jq -r -e --arg key "${key}" .$key[]) )
```
HarryWaschkeit, line 271: again stdin redirection could be used.
and maybe it is better to surround `${preserve[@]}` with double quotes just for the case of strange filenames.

recipes-omnect/initrdscripts/omnect-os-initramfs/omnect-device-service-setup
```diff
@@ -20,59 +25,44 @@ get_id() {
...
+            mv /tmp/fsck.json.tmp /tmp/fsck.json
             set_bootloader_env_var omnect_fsck_${i} ""
         fi
     done
 
-    run_cmd chown ${omnect_device_service_user}:${omnect_device_service_group} ${omnect_run_dir}/fsck.json || return 1
-    run_cmd chmod 600 ${omnect_run_dir}/fsck.json || return 1
-
-    return 0
+    cat ${omnect_os_initramfs_json} | jq --argjson fsck "$(cat /tmp/fsck.json)" '.fsck=$fsck' > /tmp/omnect-os-initramfs.json
```
HarryWaschkeit, line 40: unnecessary cat

recipes-omnect/initrdscripts/omnect-os-initramfs/omnect-device-service-setup
```diff
@@ -20,59 +25,44 @@ get_id() {
...
-    if [ -n "${status}" ]; then
-        echo ${status} > ${omnect_run_dir}/factory-reset-status
-        set_bootloader_env_var ${FACTORY_RESET_STATUS_VAR}
-    fi
-
-    run_cmd chown ${omnect_device_service_user}:${omnect_device_service_group} ${omnect_run_dir}/factory-reset-status || return 1
-    run_cmd chmod 600 ${omnect_run_dir}/factory-reset-status || return 1
-
-    return 0
+    cat ${omnect_os_initramfs_json} | jq --argjson factory_reset "$(cat /tmp/factory-reset.json)" '."factory-reset"=$factory_reset' > /tmp/omnect-os-initramfs.json
```
HarryWaschkeit, line 47: unnecessary cat

README.md
```diff
@@ -366,40 +367,69 @@ There is also the custom wipe mode. This mode provides the possibility to addres
...
+This example preserves the `bash_history` of the users `omnect` and `root`.
 
-The list of path names is separated by the character `;` and is enclosed by the `'` quotation mark.
-The factory reset is directed to the partitions `etc` and `data`.
-Therefore, path names with the following prefixes are allowed: `/etc/`, `/home/`, `/var/lib/`, `/var/log/`, `/usr/local/` and /mnt/data.
-
-In the case of an error during the backup of a file or directory part of the restore list, the whole factory reset will be aborted
+#### Factory Reset Result
+In the case of an error during the backup of a files or directories the whole factory reset will be aborted
 and the partitions `etc` and `data` remain untouched.
```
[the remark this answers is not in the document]
mlilien: is it really important? it will work with /home/ or /mnt/data/home/ ... the mechanism doesn't distinguish.
it will fail if the path you use isn't overlayed though.
do you have a proposal?

README.md
```diff
@@ -366,40 +367,69 @@ There is also the custom wipe mode. This mode provides the possibility to addres
...
+```json
+{
+  "status": 0,
+  "error": "0",
+  "paths": [
+    "/etc/omnect/factory-reset.d/",
+    "/etc/restore_file",
+    "/home/omnect/.bash_history",
+    "/home/root/.bash_history"
+  ]
```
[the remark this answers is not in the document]
mlilien: it repeats all configured paths from keys from the request matching /etc/omnect/factory-reset.json and /etc/omnect/factory-reset.d/*.json if the request contained "applications".

README.md
```diff
@@ -366,40 +367,69 @@ There is also the custom wipe mode. This mode provides the possibility to addres
...
   - 0: wipe mode supported
   - 1: wipe mode unsupported
   - 2: backup/restore failure
-- *subordinated status* (execution exit status): in case of *main status* == 0 (success)
-
-In the case of a successfully performed factory reset, the OS bootloader environment variable `factory-reset-status` is set to the value `0:0`.
+  - 3: configuration error; see "context" for details
+- `error`:  execution exit status. in case of of success == 0, if not applicaple: `-`.
+- optional: `context` on warnings or errors
+- array `paths` of preserved files or directories
```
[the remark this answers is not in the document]
mlilien: they will appear here, as long as they are configured either in /etc/omnect/factory-reset.json or /etc/omnect/factory-reset.d/*.json 
if the path doesn't exist, the factory-reset will not fail. currently we only warn in context of omnect-os-initramfs but not as content of the factory-reset result.
maybe we should refine that a little bit more @HarryWaschkeit @JanZachmann

recipes-omnect/bootloader_env/bootloader-env/bootloader_env_grub.sh
```diff
@@ -25,14 +25,14 @@ function list(){
 function set () {
     [[ ${argsc} -ne 3 ]] && help && exit 1
     local key=${1}
-    local value=${2}
-    grub-editenv ${grubenv} set ${key}=${value}
+    local value=${@:2}
+    grub-editenv ${grubenv} set "${key}"="${value}"
```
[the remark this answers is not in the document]
mlilien: yes ... try to follow the examples

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
```diff
@@ -152,41 +122,70 @@ factory_reset_backup() {
...
         msg "->does not exist; ignore..."
         return 0
     fi
 
-    cd ${FACTORY_RESET_BACKUP_DIR}/${ROOTFS_DIR}
-        # do restore, keep permissions and ownership
-        run_cmd cp --parents -av "${path_name:1}" "${ROOTFS_DIR}"
+    while [[ "${destpath}" != "${ROOTFS_DIR}" ]]; do
+        mountpoint -q ${destpath} && nested_mount=${destpath/${ROOTFS_DIR}} && break
+        destpath=${destpath%\/*}
```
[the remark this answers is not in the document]
mlilien: `destpath`is altered by the loop.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
```diff
@@ -152,41 +122,70 @@ factory_reset_backup() {
...
+    done
+
+    # do restore, keep permissions and ownership
+    if [ -n "${nested_mount}" ]; then
+        msg "nested mount: ${factory_reset_backup_dir}/${ROOTFS_DIR}${nested_mount}"
+        cd ${factory_reset_backup_dir}/${ROOTFS_DIR}${nested_mount}
+        path_name=${path_name/${nested_mount}}
```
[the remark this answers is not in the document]
mlilien: i'll test that.
what is the benefit though? we would remove the first occurrence only, which is pretty much the prefix.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
mlilien: it kinda is ... why should i omit it?

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
mlilien: true ... do you see a special benefit in doing it that way?

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
mlilien: > maybe it is better to surround ${preserve[@]} with double quotes just for the case of strange filenames.

please elaborate. if these files are added to the array correctly, this shouldn't be a problem.
can you give an example which would fail?

recipes-omnect/initrdscripts/omnect-os-initramfs/omnect-device-service-setup
mlilien: https://github.com/omnect/meta-omnect/pull/503#discussion_r1816655109

recipes-omnect/initrdscripts/omnect-os-initramfs/omnect-device-service-setup
mlilien: https://github.com/omnect/meta-omnect/pull/503#discussion_r1816655109

README.md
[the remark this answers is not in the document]
HarryWaschkeit: maybe something like "Note that home directories in example above work because they are actually located on data partition via overlay mechanism".
at least I was a bit confused about reading only about "mnt" and "data" partitions here where the example used home directories (what is a perfectly valid application of this mechanism imo), so I felt it could be worth a little note about the background.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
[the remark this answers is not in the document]
mlilien: i guess you want me to change to `destpath=$(dirname ${destpath})`? which i'll do.

please simply say that... i needed some time to understand, what could possibly a problem here.
for now i understand that there is no problem, but a readability benefit, if i adapt.

README.md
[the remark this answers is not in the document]
mlilien: so the problem is the initial documentation? is it better to remove the existing hints then?

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
HarryWaschkeit: ok, it doesn't hurt, right.
but what does `restore_list` make different from any other plain variable, e.g. like `i`, to use `[@]` here?
it's rather confusing than helping here imho, as it suggests we're dealing with an array here what is not the case.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
HarryWaschkeit: avoidance of extra process invocation could be an argument here.
but to me use of file redirection always appeared the Linux way and an unneeded pipe doesn't make things clearer, at least here the simple `< ${i}` looks better imho.
it might feel different in a longer pipe construct, in the end this is a question of preference.
my 0,02 €.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
HarryWaschkeit: parameter expansion of `${preserve[@]}` underlies standard word splitting in the for loop construct, weird elements of `preserve` array containing spaces would lead ti errors.
using `for key in "${preserve[@]}" ...` would prevent that.

mlilien: an example would be easier to understand what you want from me.
e.g. unexpected but accepted:
```
$ foo=( "1" "2" "3" )
$ foo+=( "4 5" )
$ for i in ${foo[@]}; do echo $i; done
1
2
3
4
5
$ for i in "${foo[@]}"; do echo $i; done
1
2
3
4 5

```

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
mlilien: here i could even change it to (and will so)
```
preserve_list+=( $(jq -r -e '.paths[]' ${i}) )
```
without an pipe like `|`or `<`. 
from the result it imho doesn't make a difference though.
"extra process invocation" may be an argument, is this an established rule though?

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
mlilien: i'm confused. what makes you think, we're not dealing with an array here? see line 160.

HarryWaschkeit: line 160 expands to a string containing all positional arguments starting with the 2nd and separated by spaces.
if you wanted to have an array, parentheses would be necessary like used in lines 250 and 252 for variable `preserve`, i.e. `local restore_list=("${@:2}")`

README.md
```diff
@@ -366,40 +367,69 @@ There is also the custom wipe mode. This mode provides the possibility to addres
...
 ```
+This example preserves the `bash_history` of the users `omnect` and `root`.
 
-The list of path names is separated by the character `;` and is enclosed by the `'` quotation mark.
-The factory reset is directed to the partitions `etc` and `data`.
-Therefore, path names with the following prefixes are allowed: `/etc/`, `/home/`, `/var/lib/`, `/var/log/`, `/usr/local/` and /mnt/data.
-
-In the case of an error during the backup of a file or directory part of the restore list, the whole factory reset will be aborted
+#### Factory Reset Result
+In the case of an error during the backup of a files or directories the whole factory reset will be aborted
```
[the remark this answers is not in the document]
mlilien: fixed typo.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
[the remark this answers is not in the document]
mlilien: Will be handled in a future PR.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
mlilien: Will possibly be revisited in a future PR.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
mlilien: Will be handled in a future PR.

recipes-omnect/initrdscripts/omnect-os-initramfs/factory-reset
mlilien: Will be handled in a future PR.
I'll will add a test which contains filenames with Spaces (and other not so nice Tokens - Ideas welcome),

PR #524, opened 2024-12-03
feat: build cyclonedx boms for omnect-device-service, wifi-commissioning-gatt-service

mlilien:


recipes-devtools/cargo-cyclonedx/cargo-cyclonedx_0.5.7.bb
```diff
@@ -0,0 +1,234 @@
...
+
+SUMMARY = "CycloneDX Software Bill of Materials (SBOM) for Rust Crates"
+HOMEPAGE = ""
+LICENSE = ""
+
+# includes this file if it exists but does not fail
+# this is useful for anything you may want to override from
+# what cargo-bitbake generates.
+include cargo-cyclonedx-${PV}.inc
+include cargo-cyclonedx.inc
```
HarryWaschkeit, line 234: given that we add cargo-cyclonedx.inc here wouldn't be "require" the better statement?
(opposed to cargo-cyclonedx-${PV}.inc which is no part of the repo)
or are the include statement maybe stemming from cargo bitbake and we just use the version agnostic include file already for our specific overrides?

PR #564, opened 2025-03-27
feat(genericx86): secure boot

mlilien:


mlilien:
A/B update is not working with meta-secure-core's kirkstone branch. (scarthgap works though).

PR #572, opened 2025-04-15
chore(iot-hub-device-update): update to 1.2.0

JoergZeidler:


mlilien (approved):
i suggest to change the title to `chore(iot-hub-device-update: update to 1.2.0` but decide yourself. reason is a different formatting in the resulting changelog.

PR #584, opened 2025-05-08
feat: omnect-device-service with factory-reset v3

JanZachmann:
- return `null` from initramfs in case no factory reset happened in previous boot
- bump omnect-device-service to 0.41.1 which supports
  - factory reset version 3
  - rust edition 2024


recipes-omnect/omnect-device-service/omnect-device-service_0.41.1.bb
```diff
@@ -1,16 +1,440 @@
-# generated recipes with cargo-bitbake 0.3.16 are not compatible with scarthgap per se.
-# we have to include the sha256sum for all packages
-# (output of yocto build error when missing:)
```
mlilien, line 3: only fyi: weird decission of github, to use this file as a base of renaming

PR #598, opened 2025-06-04
feat(kernel): added USB serial drivers for Prolific (PL2303) and QinHeng (CH341)

HarryWaschkeit:
This allows to use omnect devices as serial console supervisor for other devices beside the widespread FTDI based USB-to-Serial converters.


HarryWaschkeit:
fiyi @JanZachmann @JoergZeidler

conf/machine/phygate-tauri-l-imx8mm-2.extra.conf
```diff
@@ -12,6 +12,8 @@ MACHINE_FEATURES:remove = " \
     wifi \
 "
 
+MACHINE_FEATURES:append = " extra-usb-serial-devices "
+
```
HarryWaschkeit, line 16: we could also add this feature for development images only in case of security concerns.
however, I do not see any additional attack surface with that extra drivers: the only difference is that now there are two more USB devices (beside the already supported FTDI chips) which can lead to a serial device being available (but unused by default) in the system.

HarryWaschkeit: changed to make drivers generally available.

mlilien (approved):

PR #599, opened 2025-06-05
fix(networkmanager): get rid of nftables and python in rootfs

JoergZeidler:


recipes-connectivity/networkmanager/networkmanager_%.bbappend
```diff
@@ -16,3 +16,5 @@ do_install:append() {
 }
 
 ALTERNATIVE_PRIORITY[resolv-conf] = "10"
+# to get rid of nftables and python in rootfs, disable the networkmanager firewall
```
HarryWaschkeit, line 19: maybe we should state that the firewall from networkmanager is not used at all so neither iptables nor nftables are needed.
however, the hint about nftables's dependency to python and the fact that python is currently unused otherwise in our devices is still valuable imho.

JoergZeidler: done

recipes-connectivity/networkmanager/networkmanager_%.bbappend
```diff
@@ -16,3 +16,7 @@ do_install:append() {
 }
 
 ALTERNATIVE_PRIORITY[resolv-conf] = "10"
+
+# Notice: firewall from networkmanager is not used at all, so nftables or iptables settings are not needed.
+# To remove nftables and python from the root filesystem, the network manager firewall has been disabled.
+NETWORKMANAGER_FIREWALL_DEFAULT = ""
```
mlilien, line 22: - "network manager"  -> "networkmanager"/"NetworkManager"
- the networkmanager is not used to configure the firewall ... there is no explicit "network manager firewall".

JoergZeidler: comment refactored

PR #607, opened 2025-07-22
refactor(wpa_supplicant): added automatic periodic (5s) restart upon failure for 120s

HarryWaschkeit:
This is done with the intention to tackle arrakis mk4 sporadic start-up issue with Atheros chip.

### DIFF
--- a/recipes-connectivity/wpa-supplicant/wpa-supplicant/wpa_supplicant@.service
+++ b/recipes-connectivity/wpa-supplicant/wpa-supplicant/wpa_supplicant@.service
@@ -7,11 +7,15 @@ Wants=network.target
 # unfortunately this has no real benefit: https://github.com/systemd/systemd/issues/985
 ConditionPathExists=/sys/class/net/%i
 
+StartLimitIntervalSec=120
+
 # NetworkManager users will probably want the dbus version instead.
 
 [Service]
 Type=simple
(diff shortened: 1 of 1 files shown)


mlilien (approved):

recipes-connectivity/wpa-supplicant/wpa-supplicant/wpa_supplicant@.service
```diff
@@ -7,11 +7,15 @@ Wants=network.target
 # unfortunately this has no real benefit: https://github.com/systemd/systemd/issues/985
 ConditionPathExists=/sys/class/net/%i
 
+StartLimitIntervalSec=120
+
```
JoergZeidler, line 11: like ods or adu service files, we have also added StartLimitBurst

HarryWaschkeit: as far as I understand the service would now be restarted every 5s within a time frame of 120s before giving up on starting it.
I can add StartLimitBurst, but Is there any advantage of also limiting the number of restarts here?
As I didn't see any output on failed service launches I don't think so but I'm open for arguments :-)

HarryWaschkeit: ok, for the sake of uniformity I added StartLimitBurst=10

PR #637, opened 2026-03-09
refactor: kernel argument handling

mlilien:
- added script `omnect_extra_bootargs.sh` to merge `/boot/omnect_extra_bootargs_omnect` and `/boot/omnect_extra_bootargs_custom`
- adu + ods handle new bootargs in `/boot/omnect_extra_bootargs_omnect` and `/boot/omnect_extra_bootargs_custom`
- swupdate execution is restricted to group disk
- swupdate cap cap_sys_admin+ep is set to enable mount /dev/omnect/boot for raw file handler
- initramfs: /boot gets mounted with umask 0002 and gid=disk to fix swupdate permission problems when used by ods
- unified vfat nls iocharset: iso8859-1
- raspberrypi: removed uboot env var `quirks_set`

**note**: this PR enforces bootloader updates on all plattforms


recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+#todo check if boot is mounted
+# check if we're root
+[ -f /boot/extra_bootargs_omnect ] || touch /boot/extra_bootargs_omnect
+[ -f /boot/extra_bootargs_custom ] || touch /boot/extra_bootargs_custom
+
+current_bootargs=$(bootloader_env.sh get extra_bootargs)
+new_bootargs="$(cat /boot/extra_bootargs_omnect) $(cat /boot/extra_bootargs_custom)"
```
HarryWaschkeit: is there a reason to use two cat processes instead of passing both files right away?

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
```diff
@@ -0,0 +1,15 @@
...
+new_bootargs="$(echo ${new_bootargs} | awk '{$1=$1};1')" # remove possibly trailing space
+
+
+if [ "${current_bootargs}" != "${new_bootargs}" ] && [ -n "${new_bootargs}" ]; then
```
HarryWaschkeit: rather cosmetic: for empty new_bootargs checking it first skips comparison check

dynamic-layers/raspberrypi/recipes-bsp/u-boot/u-boot/omnect_env_rpi.h
```diff
@@ -1,15 +1,14 @@
...
 // ToDo: as soon we can get rid of boot scripts, we should create different u-boot envs for images vs. test-images
 /* Attention: if vars are already part of CONFIG_EXTRA_ENV_SETTINGS
  * this doesnt work; these vars are shown correctly via userland fw_printenv,
  * but u-boot takes the default env value
- * This means in this case extra-bootargs and quirks_set may not be preset by omnect_env*.h
+ * This means in this case quirks_set may not be preset by omnect_env*.h
  */
 #define OMNECT_REQUIRED_WRITEABLE_ENV_FLAGS_MACHINE "," \
-    "extra-bootargs:sw," \
     "quirks_set:bw"
```
mlilien: TODO: do we still need this?

recipes-kernel/linux/files/vfat.cfg
```diff
@@ -0,0 +1 @@
+CONFIG_FAT_DEFAULT_IOCHARSET="iso8859-1"
```
HarryWaschkeit: I think it is better to have an EOL

recipes-bsp/u-boot/u-boot/omnect_env.env
HarryWaschkeit: indentation looks weird from here; it's kind of a tab-8 / tab-4 mix beginning with the `if` statement

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
```diff
@@ -348,3 +380,14 @@ flash_mode_setup_network() {
...
+# get ID field from group file
+get_id() {
+    local name="$1"
+    local file="$2"
+    local line=""
+
+    line=$(grep -E -e "^${name}:" ${file})
+    if [ -z "${line}" ]; then echo ""; return; fi
+    echo $line | awk -F: '{print $3}'
+}
```
HarryWaschkeit: EOL?

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
```diff
@@ -0,0 +1,46 @@
...
+[ -f /boot/omnect_extra_bootargs_custom ] || touch /boot/omnect_extra_bootargs_custom
+
+current_bootargs=$(bootloader_env.sh get omnect_extra_bootargs || true)
+new_bootargs="$(cat /boot/omnect_extra_bootargs_omnect) $(cat /boot/omnect_extra_bootargs_custom)"
+new_bootargs="$(echo ${new_bootargs} | awk '{$1=$1};1')" # remove possibly trailing space
+
+function help() {
+    echo "usage:"
+    echo "bootloader_env.sh command key [value]"
+    echo "    command: {get_current,get_new,set}"
```
HarryWaschkeit: looks a bit awkward but would always yield all commands as in `${commands}`:

`echo "    command: {$(IFS=,; set -- "${commands[@]}"; echo "${*}")}"`

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
```diff
@@ -138,12 +138,44 @@ setup_etc_from_factory()
...
+        msg "first boot detected"
         run_cmd mkdir -p "${rootfs_dir}/mnt/etc/upper"
         run_cmd mkdir -p "${rootfs_dir}/mnt/etc/work"
         if [ -d "${rootfs_dir}/mnt/factory/etc" ]; then
             msg "setup etc from factory partition"
             run_cmd cp -av "${rootfs_dir}/mnt/factory/etc/." "${rootfs_dir}/mnt/etc/upper"
         fi
+
+        # below has nothing to do with etc setup, but we only want to do this after flash-mode-1
+        if [[ -e /run/factory-reset ]]; then
```
HarryWaschkeit: in swupdate scripts the used test is `-f`; is this intentionally different here?

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
mlilien: the space between the results.
will change to `"$(< /boot/extra_bootargs_omnect) $(< /boot/extra_bootargs_custom)"`

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
mlilien: no, i'll adapt

README.md
mlilien: add `xattr`

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
mlilien: done

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
mlilien: refactored, but in another way

recipes-kernel/linux/files/vfat.cfg
mlilien: yep, my ide does save with EOL ... seems i didn't use it

mlilien: done

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
mlilien: done

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
mlilien: example doesn't work, but i agree. done.

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
mlilien: done

README.md
mlilien: done

recipes-bsp/u-boot/u-boot/omnect_env.env
mlilien: correct line? i don't see what you mean.

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -792,14 +787,44 @@ ApplyUpdate() {
 
     echo "Applying." >> "${log_file}"
 
-    if [ -f "/tmp/omnect-bootloader-update" ]; then
-        bootloader_env.sh set omnect_os_bootpart $update_part
+    # bootargs handling
+    # normally we would use omnect_extra_bootargs.sh here, but it isn't available prior to 5.0.16
```
HarryWaschkeit: it is clear that the script is not yet there pre 5.0.16, but:
- shouldn't we use it if is available?
- or do we have here some "FIXME" for post 5.0.15?

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
```diff
@@ -719,24 +719,19 @@ InstallUpdate() {
...
-                            ret_val=$?
-                        fi
-                    else
-                        if [ -f "/tmp/omnect-bootloader-update-not-necessary" ]; then
-                            ret_val=0
-                            rm /tmp/omnect-bootloader-update-not-necessary
-                        fi
+                    if [ ${ret_val} -ne 0 ] && [ -f "/tmp/omnect-bootloader-update-not-necessary" ]; then
+                        ret_val=0
+                        rm /tmp/omnect-bootloader-update-not-necessary
```
HarryWaschkeit, line 734: line 727/729 already removes that file, so unless I'm missing something this block will never be used.

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
```diff
@@ -138,12 +138,44 @@ setup_etc_from_factory()
...
+            mkdir -p ${ROOTFS_DIR}/boot
+            check_fs "boot"
+            # Dangerous: This function gets potentially called when /dev/omnect/* doesn't exist yet (caused by the init script system)
+            # Currently it is safe
+            mount -t vfat /dev/omnect/boot ${ROOTFS_DIR}/boot && msg "mounted ${ROOTFS_DIR}/boot"  || return 1
+            mounted=1
+        fi
+        local current_bootargs="$(get_bootloader_env_var "omnect_extra_bootargs")"
+        local new_bootargs="$(cat ${ROOTFS_DIR}/boot/omnect_extra_bootargs_omnect) $(cat ${ROOTFS_DIR}/boot/omnect_extra_bootargs_custom)"
+        new_bootargs="$(echo ${new_bootargs} | awk '{$1=$1};1')" # remove possibly trailing space
```
HarryWaschkeit, line 168: imho we should use here the same logic as now in omnect_extra_bootargs.sh:
```suggestion
        new_bootargs="$(< /boot/omnect_extra_bootargs_omnect) $(< /boot/omnect_extra_bootargs_custom)"
```
(beware of TAB vs. SPACE in suggestion, I only copy&pasted from browser)

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
```diff
@@ -138,12 +138,44 @@ setup_etc_from_factory()
...
+        msg "current_bootargs=${current_bootargs} new_bootargs=${new_bootargs}"
+        if [[ "${current_bootargs}" != "${new_bootargs}" ]]; then
+            set_bootloader_env_var "omnect_extra_bootargs" "${new_bootargs}"
+            msg "reboot by omnect_extra_bootargs handling on first boot"
+            sync
+            umount ${ROOTFS_DIR}/mnt/cert
+            umount ${ROOTFS_DIR}/mnt/etc
+            run_cmd reboot -f
```
HarryWaschkeit, line 176: do we want to have this reboot visible be reboot reasons?
I don't think so as it's no full boot yet, but wanted to bring up that question nevertheless.

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
```diff
@@ -348,3 +380,14 @@ flash_mode_setup_network() {
...
+
+# get ID field from group file
+get_id() {
+    local name="$1"
+    local file="$2"
+    local line=""
+
+    line=$(grep -E -e "^${name}:" ${file})
+    if [ -z "${line}" ]; then echo ""; return; fi
+    echo $line | awk -F: '{print $3}'
```
HarryWaschkeit: why the extra check for an empty string?
wouldn't awk print exactly the same in that case?

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
```diff
@@ -0,0 +1,46 @@
...
+new_bootargs="$(echo ${new_bootargs} | awk '{$1=$1};1')" # remove possibly trailing space
+
+function help() {
+    echo "usage:"
+    echo "bootloader_exta_bootargs.sh command key [value]"
+    echo "    command: {$(IFS=,; echo "${commands[*]}")}"
+}
+
+function get_current() {
+  echo ${current_bootargs}
```
HarryWaschkeit: imho we should better put the variable into double quotes to avoid unexpected interpretation of semicolons or other characters interpreted by the shell.

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
```diff
@@ -0,0 +1,46 @@
...
+}
+
+function get_new() {
+  echo ${new_bootargs}
```
HarryWaschkeit: same as here: https://github.com/omnect/meta-omnect/pull/637/changes#r3080875296

README.md
```diff
@@ -201,6 +201,16 @@ See [omnect-cli iot-hub-device-update configuration](https://github.com/omnect/o
 ### Set `iot-identity-service` configuration
 See [omnect-cli iot-identity-service configuration](https://github.com/omnect/omnect-cli/blob/main/README.md#identity-configuration).
 
+### Set custom kernel parameter
+To set custom kernel parameter, e.g. `foo=bar` edit `/boot/omnect_extra_bootargs_custom` and apply:
```
HarryWaschkeit: nitpick: the code below doesn't show how to edit but creates the file from scratch.

```suggestion
To set custom kernel parameters, you need to write them to file `/boot/omnect_extra_bootargs_custom` (single line or multiple lines) on the device and apply it, e.g. for setting `foo=bar` you could do this:
```

README.md
```diff
@@ -201,6 +201,16 @@ See [omnect-cli iot-hub-device-update configuration](https://github.com/omnect/o
...
 
+### Set custom kernel parameter
+To set custom kernel parameter, e.g. `foo=bar` edit `/boot/omnect_extra_bootargs_custom` and apply:
+```sh
+sudo -i
+echo -n "foo=bar" > /boot/omnect_extra_bootargs_custom
+omnect_extra_bootargs.sh set
+reboot
+```
+Afterwards the kernel cmdline includes the parameter `foo=bar`.
```
HarryWaschkeit: nitpick, although it should be clear for everybody seeing a Linux system not the first time:

```suggestion
From next system start on the kernel cmdline will include the parameter `foo=bar`.
```

recipes-bsp/u-boot/u-boot/omnect_env.env
HarryWaschkeit: the starting `if` for that `else` in line 18 isn't indented but the `else` as well as the `fi` below.
that looks strange to me, as does the rest of indenting: the block under the `if` (line 18) is indented by 12 spaces whereas the `if ... else ... fi` within the `else` block (lines 28-45) is all indented by 4 spaces, the conditionals and again the blocks they contain.
and having that block extra-indented would be inconsistent with other continued lines preceding this complete block.

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
mlilien: it is possibly (re)created when doing 
`swupdate -v -i "${image_file}" -k "${public_key_file}" -e stable,bootloader &>> "${swupdate_log_file}"`

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
mlilien: There will be no script if you want to update e.g. from 5.0.9 to 5.0.17.
And imho it's not a FIXME it's working without it. should i remove the comment?

recipes-bsp/u-boot/u-boot/omnect_env.env
mlilien: would've been easier to understand if you said that this is a HEAD problem, not a problem of this PR per se. done.

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
mlilien: wrong line? i guess you want me to get rid of the `cat` call. done

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
mlilien: @JanZachmann

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
mlilien: this block is not new and was moved ... i agree though. done.

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
mlilien: done

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_extra_bootargs.sh
mlilien: done

README.md
mlilien: done, but rephrased.

README.md
mlilien: done, but rephrased.

recipes-omnect/initrdscripts/omnect-os-initramfs/common-sh
mlilien: we only should reach this on a real first boot (e.g. not on a factory-reset). 
and even then only if there were bootargs injected, e.g. by relabel process.
so using a reboot reason would possibly irritate the user on a first boot.

HarryWaschkeit: I agree concerning possible user confusion: afaics a user can not know about that extra reboot (no network initialized yet, so even indirectly monitoring that via ICMP packets wouldn't be possible).
so as I said, I don't see why it should be reflected, if it were it possibly even could confuse reboot reason detection logic later on when the system comes up.
just wanted this having discussed to ensure I didn't overlook anything.

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/swupdate_handler_v2_grub.sh
HarryWaschkeit: ok, I see we cannot get rid of the manual handling (opposed to using omnect_extra_bootargs.sh) as long as we have to support updates from pre-5.0.16 releases.

nevertheless, at first it seemed more reasonable to me to let it do the script if it is there.

but reconsidering that we have to ensure the update path from pre-5.0.16 anyway for the time being, changes from a possibly evolving  `omnect_extra_bootargs.sh` script need to be reflected here anyway; so it works rather as a safe guard to keep that update path covered.

maybe adapting the comment to something like:
```suggestion
    # normally we would use omnect_extra_bootargs.sh here, but as long as we have to support updates from releases prior to 5.0.16 where it isn't available, the logic has to be implemented here directly.
```

HarryWaschkeit (commented):
pls have a look at https://github.com/omnect/meta-omnect/pull/637/changes#r3085499451

HarryWaschkeit (approved):

PR #640, opened 2026-04-07
refactor: removed cramfs and added fstrim systemd service & timer files

HarryWaschkeit:
Filesystem `cramfs` was never used so get rid of related utilities in image.

While utility `fstrim` was already part of the images it wasn't used yet due to missing service and timer files. Added these files to prevent SSD degradation over time because storage devices get never triggered their internal house keeping of used sectors.

### DIFF
--- a/recipes-core/util-linux/util-linux_%.bbappend
+++ b/recipes-core/util-linux/util-linux_%.bbappend
@@ -1,14 +1,12 @@
 SYSTEMD_AUTO_ENABLE:omnect-os:util-linux-fstrim = "enable"
-#PACKAGECONFIG[systemd] += "--with-systemd --with-systemdsystemunitdir=${systemd_system_unitdir}"
-#PACKAGECONFIG[systemd] = "--with-systemd --with-systemdsystemunitdir=${systemd_system_unitdir}"
-# we don't need cramfs stuff so remove it from deploy dir
+
 do_install:append() {
     # install fstrim's systemd service and timer files
     install -d -m 0755  ${D}${systemd_system_unitdir}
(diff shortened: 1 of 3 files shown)


HarryWaschkeit:
@JanZachmann @JoergZeidler fyi

recipes-core/util-linux/util-linux_%.bbappend
```diff
@@ -0,0 +1,19 @@
+SYSTEMD_AUTO_ENABLE:omnect-os:util-linux-fstrim = "enable"
+#PACKAGECONFIG[systemd] += "--with-systemd --with-systemdsystemunitdir=${systemd_system_unitdir}"
+#PACKAGECONFIG[systemd] = "--with-systemd --with-systemdsystemunitdir=${systemd_system_unitdir}"
```
mlilien: remove comments

recipes-core/util-linux/util-linux_%.bbappend
```diff
@@ -0,0 +1,19 @@
...
+# we don't need cramfs stuff so remove it from deploy dir
```
mlilien: reading this, i expected we only do this in `do_install`. imho, move the comment to line 11.

recipes-core/util-linux/util-linux_%.bbappend
HarryWaschkeit: done

recipes-core/util-linux/util-linux_%.bbappend
HarryWaschkeit: done

mlilien (approved):

PR #649, opened 2026-04-29
fix(initramfs): prevent stuck system after flash-mode-3 errors

HarryWaschkeit:
If for whatever reason not all variables required for flash mode 3 got set correctly, the flash-mode-3 init script could fail leaving a system behind with a shell in the initramfs.

Even worse, it was possible that not even a power-cycle could revive the system, because the script left the flash mode operation panding by not cleaning up the environment.

Admittedly, flash modes are generally devlopment-only feature, but possibly having to manually attend a device to sort things out could be rather inconvenient.

This change improves the situation so that at least after having failed a flash mode (3) operation a power cycle brings up the system again as it was before.

### DIFF
--- a/recipes-omnect/initrdscripts/omnect-os-initramfs/flash-mode-3
+++ b/recipes-omnect/initrdscripts/omnect-os-initramfs/flash-mode-3
@@ -53,8 +53,18 @@ run_bmap_tool() {
 
 # deploy WIC image via network to (same) active block device
 run_flash_mode_3() {
+    # we know that flash mode 3 is active or otherwise flash_mode_3_enabled()
+    # would have failed.
+    # in order to avoid to endlessly fall into this situation if some error
+    # happens below, unset flash-mode variable here right now and also clear
+    # other variables right after having been read
+    set_bootloader_env_var flash-mode
+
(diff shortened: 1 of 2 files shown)


recipes-omnect/initrdscripts/omnect-os-initramfs/flash-mode-3
```diff
@@ -53,8 +53,18 @@ run_bmap_tool() {
 
 # deploy WIC image via network to (same) active block device
 run_flash_mode_3() {
+    # we know that flash mode 3 is active or otherwise flash_mode_3_enabled()
+    # would have failed.
+    # in order to avoid to endlessly fall into this situation if some error
+    # happens below, unset flash-mode variable here right now and also clear
+    # other variables right after having been read
```
mlilien, line 60: only fyi: i don't need this comment

mlilien (approved):

recipes-omnect/initrdscripts/omnect-os-initramfs/flash-mode-3
HarryWaschkeit: I'll keep it, this file is about to die anyway with our rust based init.

PR #650, opened 2026-04-29
refactor(grub): increased environment size to 2kB

HarryWaschkeit:
Why: feature `flash-mode-3` allows to flash a device with an externally available (retrievable via HTTP[S]) image file.
Two environment variables - beside the `flash-mode` variable which chooses among the available flash modes - control what is to be flashed:

- one contains the URL to a `wic.xz` file containing the image to be flashed to the boot device
- the other one holding the SHA256sum for the retrieved image file for verification

**NOTE:** have a look at chapter "Flash Mode 3" in README.md file for full details.

Mainly the URL can be rather lengthy, possibly exceeding the standard environmrent size, together with all other variables normally set.

For this an possibly other future use cases the `grubenv` got extended.

**NOTE:** grub documentation actually states to only use `grub-editenv <file> create` to create an ampty einvironment file, and also for any modification the `grub-editenv` program shall be used. However, the `grub` source code currently does not make any assumptions about the file size.
The only check implemented to validate en environment file is whether it starts with the right signature (`# GRUB Environment Block`). Otherwise, free space to set new variables is the complete end region that is filled exclusively with `#` characters: the first newline that is detected marks the end of the stored variable settings where any line starting with a `#` character is treated as comment, hence ignored during variable/value pair parsing.


recipes-bsp/grub/grub-efi_%.bbappend
```diff
@@ -15,4 +18,10 @@ do_compile:append:class-target() {
     grub-script-check ${WORKDIR}/boot-menu.inc
 }
 
+do_install:append() {
+    # we need to replace the standard grubenv file with our own one, because
+    # we use a bigger file
+    install -m 0644 ${WORKDIR}/grubenv ${D}${EFI_BOOT_PATH}/grubenv
```
mlilien, line 24: imho that is not necessary. grubenv is copied when creating the wic image. did this really change anything?

HarryWaschkeit: it did, without that a freshly generated env file would be used in image instead of ours.

mlilien: ok, because we append after meta-secure-core/meta-efi-secure-boot/recipes-bsp/grub/grub-efi-efi-secure-boot.inc 

please do this for class-target()
`do_install:append:class-target()`

HarryWaschkeit: yes, for the record the code line of interest: https://github.com/jiazhang0/meta-secure-core/blob/59ae6f8d39862eaa7f250add4b5d1b71e8aeda4c/meta-efi-secure-boot/recipes-bsp/grub/grub-efi-efi-secure-boot.inc#L137

changed to `class-target` type append

mlilien (approved):

PR #652, opened 2026-05-11
fix(initramfs): mount error and console messages on first boot

HarryWaschkeit:


recipes-omnect/initrdscripts/omnect-os-initramfs/fs-mount
```diff
@@ -18,7 +17,9 @@ run_fs_mount()
 
     # mount rootfs
     # before we mount rootfs, make sure boot is unmounted
-    umount ${ROOTFS_DIR}/boot || true
+    if mountpoint -q ${ROOTFS_DIR}/boot; then
```
mlilien, line 20: we had this test in the past, and it will work here.
the reason i removed this test  was: it doesn't work if you mounted rootfs after boot. this will hide the mountpoint from boot. 
no need to change, only reasoning why it was the way it is.

mlilien (approved):

mlilien (approved):

PR #654, opened 2026-05-26
refactor!: unified extra kernel args and handling for pstore (rpi & tauri)

HarryWaschkeit:
Previously, the CMDLINE mechanism was used to set suitable kernel arguments for reboot reason (feature pstore) handling what is actually the standard way in the Raspberry Pi BSP.

However, the standard omnect OS way is to use the APPEND mechanism which gets applied to arbitrary bootloaders.

Moreover, for tauri there was an unnecessary ramoops setting (use of ECC) part of the kernel arguments, because it is already part of the device tree add-on that was introduced for reboot reason support.

BREAKING CHANGE: this forces a boot loader update on Raspberry Pi 4.

### DIFF
--- a/conf/machine/include/rpi-pstore.inc
+++ b/conf/machine/include/rpi-pstore.inc
@@ -1,1 +1,1 @@
-CMDLINE:append:pn-rpi-cmdline:omnect_pstore = " ramoops.dump_oops=1 printk.always_kmsg_dump=Y crash_kexec_post_notifiers=Y"
+APPEND:append:omnect_pstore = " ramoops.dump_oops=1 ramoops.ecc=1 printk.always_kmsg_dump=Y crash_kexec_post_notifiers=Y"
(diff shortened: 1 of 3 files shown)


conf/machine/include/rpi-pstore.inc
```diff
@@ -1 +1 @@
-CMDLINE:append:pn-rpi-cmdline:omnect_pstore = " ramoops.dump_oops=1 printk.always_kmsg_dump=Y crash_kexec_post_notifiers=Y"
+APPEND:append:omnect_pstore = " ramoops.dump_oops=1 ramoops.ecc=1 printk.always_kmsg_dump=Y crash_kexec_post_notifiers=Y"
```
mlilien: is the additional `ramoops.ecc` worth a mentioning in the description?

mlilien (commented):
please add to the description that this PR enforces a rpi bootloader update

conf/machine/include/rpi-pstore.inc
HarryWaschkeit: good catch, I'll re-test without it, because it should actually be unnecessary while part of the device tree overlay for pstore feature.

HarryWaschkeit: as expected it works also without that parameter - removed.

mlilien (approved):

PR #655, opened 2026-06-01
chore(kas): updated meta-swupdate

HarryWaschkeit:


JoergZeidler (approved):

kas/distro/omnect-os.yaml
```diff
@@ -34,7 +34,7 @@ repos:
   ext/meta-swupdate:
     url: "https://github.com/sbabic/meta-swupdate.git"
     branch: "scarthgap"
-    commit: "81f4faa406e70dfa514e2f7642f97df33d3b84ae"
+    commit: "a329ea9f9ef4eeb316f2da2dba4a9c1bcb106899"
```
mlilien, line 37: wrong rev: should be cb709b865a56be48cb17cbad149b69f793f3888b.
why did we update anyway?

HarryWaschkeit: will be fixed with 5.0.18

PR #657, opened 2026-06-15
fix(dependency-track): allow CPE version splitting for SBOM generation

HarryWaschkeit:
Component versions in CPE entries of CVEs are used differently with respect to suffixes like `p2`, `rc1` and so on:
- for most components the full version (i.e. including any suffixes) is written to the CPE "version" field and CPE field "update" is set to `*`
- some components however use the update field for suffixes, using `-` if none is present

This cannot be handled generically, so a mechanism was introduced to specify the split handling for respective components.

See classes/dependency-track.bbclass for a detailed description.

### DIFF
--- a/classes/dependency-track.bbclass
+++ b/classes/dependency-track.bbclass
@@ -7,14 +7,14 @@
 CVE_PRODUCT ??= "${BPN}"
 CVE_VERSION ??= "${PV}"
 
-# There are differences in how component versions are hendled by their vendors
+# There are differneces in how component versions are handled by their vendors
 # with respect to CPE fields 'version' and 'update'::
 #  - most of them seem to put the whole version number into the version field
 #  - some split the version string into pure version number and a potential
 #    suffix like 'beta-1', 'rc1' or 'p1'
 # In the past we used first variant for all components, but one day a sudo CVE
 # hit the developemer in charge and revealed that projects appeared to be
-# unaffected by it although the actually are.
+# unaffected by it although they actually are.
 # The cause lies in version number comparison between CVE's CPE entries and
 # our generated SBOM for single versions: they need to fully match.
 # Concrete values were ...
@@ -28,7 +28,7 @@ CVE_VERSION ??= "${PV}"
 # The full version in the SBOM's version field didn't match the complete
 # version specification as contained in the CVE - split into version plus
 # update -, so the CVE was ignored.
-# From the CPE standard's point of view there is no right or wron here, so we
+# From the CPE standard's point of view there is no right or wrong here, so we
 # need to be prepared to deal with both situations.
 # The solution is: keep SBOM generation for components as in the past unless
 # an indicator - new variable OMNECT_CVE_VERSION_SPLIT, to be set component
@@ -96,7 +96,6 @@ python do_dependencytrack_collect() {
             # we need to use the full version from the cpe string because it
             # is already cleaned from any '+git' suffix
             full_ver = cpe_split[5]
-            #m = re.search('^([0-9]+(?:[.][0-9]+)*)([-_a-zA-Z][-_a-zA-Z0-9]+)?$', full_ver)
             m = re.search('^([0-9]+(?:[.][0-9]+)*)([-+_a-zA-Z]+.*)?$', full_ver)
             bb.debug(1, "component[split]: {} / {} ({}) - m: {}".format(name, full_ver, version, m))
             bb.debug(1, "component[split]: cpe {}". format(cpe))
--- a/classes/dependency-track.bbclass
+++ b/classes/dependency-track.bbclass
@@ -98,7 +98,7 @@ python do_dependencytrack_collect() {
             full_ver = cpe_split[5]
             m = re.search('^([0-9]+(?:[.][0-9]+)*)([-+_a-zA-Z]+.*)?$', full_ver)
             if m == None:
-                bb.error("component {}: version {} cannot be split".format(name, full_ver))
+                bb.fatal("component {}: version {} cannot be split".format(name, full_ver))
 
             bb.debug(1, "component[split]: {} / {} ({}) - m: {}".format(name, full_ver, version, m))
             bb.debug(1, "component[split]: cpe {}". format(cpe))
--- a/classes/dependency-track.bbclass
+++ b/classes/dependency-track.bbclass
@@ -7,6 +7,35 @@
 CVE_PRODUCT ??= "${BPN}"
 CVE_VERSION ??= "${PV}"
 
+# There are differences in how component versions are hendled by their vendors
+# with respect to CPE fields 'version' and 'update'::
+#  - most of them seem to put the whole version number into the version field
+#  - some split the version string into pure version number and a potential
+#    suffix like 'beta-1', 'rc1' or 'p1'
+# In the past we used first variant for all components, but one day a sudo CVE
+# hit the developemer in charge and revealed that projects appeared to be
+# unaffected by it although the actually are.
+# The cause lies in version number comparison between CVE's CPE entries and
+# our generated SBOM for single versions: they need to fully match.
+# Concrete values were ...
+#  - in the CVE:
+#    cpe:2.3:a:sudo_project:sudo:1.9.17:-:*:*:*:*:*:*
+#    cpe:2.3:a:sudo_project:sudo:*:*:*:*:*:*:*:* (<1.9.17)
+#    cpe:2.3:a:sudo_project:sudo:1.9.17:p1:*:*:*:*:*:*
+#    cpe:2.3:a:sudo_project:sudo:1.9.17:p2:*:*:*:*:*:*
+# in our SBOM:
+#    cpe:2.3:a:*:sudo:1.9.17p1:*:*:*:*:*:*:*
+# The full version in the SBOM's version field didn't match the complete
+# version specification as contained in the CVE - split into version plus
+# update -, so the CVE was ignored.
+# From the CPE standard's point of view there is no right or wron here, so we
+# need to be prepared to deal with both situations.
+# The solution is: keep SBOM generation for components as in the past unless
+# an indicator - new variable OMNECT_CVE_VERSION_SPLIT, to be set component
+# specific in omnect-os-cve.conf file - states otherwise ie., sets this
+# variable to 1.
+OMNECT_CVE_VERSION_SPLIT ??= "0"
+
 DEPENDENCYTRACK_DIR ??= "${DEPLOY_DIR}/dependency-track"
 DEPENDENCYTRACK_SBOM ??= "${DEPENDENCYTRACK_DIR}/bom.json"
 DEPENDENCYTRACK_TMP ??= "${TMPDIR}/dependency-track"
@@ -38,6 +67,7 @@ do_dependencytrack_init[eventmask] = "bb.event.BuildStarted"
 python do_dependencytrack_collect() {
     import json
     import oe.cve_check
+    import re
     from pathlib import Path
 
     # omnect specific; we only want packages for the target in our sbom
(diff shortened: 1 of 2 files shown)


classes/dependency-track.bbclass
```diff
@@ -7,6 +7,35 @@
 CVE_PRODUCT ??= "${BPN}"
 CVE_VERSION ??= "${PV}"
 
+# There are differences in how component versions are hendled by their vendors
```
mlilien: handled

classes/dependency-track.bbclass
```diff
@@ -7,6 +7,35 @@
...
+# with respect to CPE fields 'version' and 'update'::
+#  - most of them seem to put the whole version number into the version field
+#  - some split the version string into pure version number and a potential
+#    suffix like 'beta-1', 'rc1' or 'p1'
+# In the past we used first variant for all components, but one day a sudo CVE
+# hit the developemer in charge and revealed that projects appeared to be
+# unaffected by it although the actually are.
```
mlilien: they?

classes/dependency-track.bbclass
```diff
@@ -7,6 +7,35 @@
...
+#    cpe:2.3:a:sudo_project:sudo:1.9.17:-:*:*:*:*:*:*
+#    cpe:2.3:a:sudo_project:sudo:*:*:*:*:*:*:*:* (<1.9.17)
+#    cpe:2.3:a:sudo_project:sudo:1.9.17:p1:*:*:*:*:*:*
+#    cpe:2.3:a:sudo_project:sudo:1.9.17:p2:*:*:*:*:*:*
+# in our SBOM:
+#    cpe:2.3:a:*:sudo:1.9.17p1:*:*:*:*:*:*:*
+# The full version in the SBOM's version field didn't match the complete
+# version specification as contained in the CVE - split into version plus
+# update -, so the CVE was ignored.
+# From the CPE standard's point of view there is no right or wron here, so we
```
mlilien: wrong

classes/dependency-track.bbclass
```diff
@@ -61,7 +92,30 @@ python do_dependencytrack_collect() {
 
         # recompile cpe: scarthgap sets "*" for type, kirkstone did set "a"
         cpe_split[2] = "a"
+        if split_version != "0":
+            # we need to use the full version from the cpe string because it
+            # is already cleaned from any '+git' suffix
+            full_ver = cpe_split[5]
+            #m = re.search('^([0-9]+(?:[.][0-9]+)*)([-_a-zA-Z][-_a-zA-Z0-9]+)?$', full_ver)
```
mlilien: rm

classes/dependency-track.bbclass
mlilien: btw: can you change the title to something more meaningful, like "split cpe version" ?

HarryWaschkeit: fixed & changed

classes/dependency-track.bbclass
HarryWaschkeit: fixed

classes/dependency-track.bbclass
HarryWaschkeit: fixed

classes/dependency-track.bbclass
HarryWaschkeit: removed, also refactored a bit

classes/dependency-track.bbclass
HarryWaschkeit: not fully happy with the (hopefully) more informative title, though, now the error that was fixed is no longer part of it (but title is already rather long) ...

classes/dependency-track.bbclass
```diff
@@ -61,7 +92,26 @@ python do_dependencytrack_collect() {
...
         cpe_split[2] = "a"
+        if split_version != "0":
+            # we need to use the full version from the cpe string because it
+            # is already cleaned from any '+git' suffix
+            full_ver = cpe_split[5]
+            m = re.search('^([0-9]+(?:[.][0-9]+)*)([-+_a-zA-Z]+.*)?$', full_ver)
+            bb.debug(1, "component[split]: {} / {} ({}) - m: {}".format(name, full_ver, version, m))
+            bb.debug(1, "component[split]: cpe {}". format(cpe))
+            cpe_split[5] = m[1]
+            if len(m.groups()) > 1:
```
mlilien: claude:

  🔴 Critical — len(m.groups()) > 1 is constant-true; no-suffix versions crash the build
  
  classes/dependency-track.bbclass, in do_dependencytrack_collect:

  m = re.search('^([0-9]+(?:[.][0-9]+)*)([-+_a-zA-Z]+.*)?$', full_ver)
  ...
  cpe_split[5] = m[1]
  if len(m.groups()) > 1:      # <-- always True
      cpe_split[6] = m[2]
  else:
      cpe_split[6] = '-'       # <-- dead code, never reached

  The regex has two capturing groups, so m.groups() is always a 2-tuple and len(...) is always 2. The condition is constant-true, and the else branch (the documented "use - if none present" case) is dead code.

  When a split-enabled component is at a version without a suffix (e.g. sudo 1.9.17, not 1.9.17p1), group 2 does not match → m[2] is None → cpe_split[6] = None → ":".join(cpe_split) raises:

  TypeError: sequence item 6: expected str instance, NoneType found

  That fails do_dependencytrack_collect for that recipe, i.e. a build failure. This is a realistic trigger — sudo's base point releases (1.9.17, 1.9.18, …) carry no pN suffix, and the very CVE example in your comment (...:sudo:1.9.17:-:...) is exactly the no-suffix case the code can't produce.

  Reproduced:
  1.9.17  -> groups=('1.9.17', None)  len(groups)=2
  cpe_split[6] = None   (intended: '-')
  JOIN CRASH -> sequence item 6: expected str instance, NoneType found
  
  Fix: test whether group 2 actually matched, not the group count:
  if m[2] is not None:        # or: if m.group(2):
      cpe_split[6] = m[2]
  else:
      cpe_split[6] = '-'

classes/dependency-track.bbclass
```diff
@@ -61,7 +92,26 @@ python do_dependencytrack_collect() {
 
         # recompile cpe: scarthgap sets "*" for type, kirkstone did set "a"
         cpe_split[2] = "a"
+        if split_version != "0":
+            # we need to use the full version from the cpe string because it
+            # is already cleaned from any '+git' suffix
+            full_ver = cpe_split[5]
+            m = re.search('^([0-9]+(?:[.][0-9]+)*)([-+_a-zA-Z]+.*)?$', full_ver)
```
mlilien, line 99: claude:
 🟠 Medium — m is None is unguarded; non-numeric-leading versions crash
  
  The regex anchors on a leading digit (^[0-9]+). Any version that doesn't start with a digit (a v1.2.3 style PV, a git-describe string, a date tag, an epoch-prefixed value) makes re.search return None, and the next line cpe_split[5] = m[1] raises 'NoneType' object is not subscriptable — again a task/build failure.

  Reproduced:
  v1.2.3 -> m is None -> m[1] CRASH: 'NoneType' object is not subscriptable

  sudo/openssh happen to be numeric today, but OMNECT_CVE_VERSION_SPLIT is a general per-recipe knob; the next person who sets it on a component with an unusual PV gets a hard failure. Guard it:
  if m is None:
      bb.warn("OMNECT_CVE_VERSION_SPLIT set for %s but version %r is unparseable; leaving CPE unsplit" % (name, full_ver))
  else:
      ...

classes/dependency-track.bbclass
```diff
@@ -7,6 +7,35 @@
 CVE_PRODUCT ??= "${BPN}"
 CVE_VERSION ??= "${PV}"
 
+# There are differneces in how component versions are handled by their vendors
```
JoergZeidler: Typo: `differneces` → `differences`

classes/dependency-track.bbclass
```diff
@@ -7,6 +7,35 @@
...
+# with respect to CPE fields 'version' and 'update'::
```
JoergZeidler: Double colon at end of line: `::` → `:`

classes/dependency-track.bbclass
```diff
@@ -7,6 +7,35 @@
...
+#  - most of them seem to put the whole version number into the version field
+#  - some split the version string into pure version number and a potential
+#    suffix like 'beta-1', 'rc1' or 'p1'
+# In the past we used first variant for all components, but one day a sudo CVE
+# hit the developemer in charge and revealed that projects appeared to be
```
JoergZeidler: Typo: `developemer` → `developer`

classes/dependency-track.bbclass
```diff
@@ -61,7 +92,26 @@ python do_dependencytrack_collect() {
...
+            full_ver = cpe_split[5]
+            m = re.search('^([0-9]+(?:[.][0-9]+)*)([-+_a-zA-Z]+.*)?$', full_ver)
+            bb.debug(1, "component[split]: {} / {} ({}) - m: {}".format(name, full_ver, version, m))
+            bb.debug(1, "component[split]: cpe {}". format(cpe))
+            cpe_split[5] = m[1]
+            if len(m.groups()) > 1:
+                cpe_split[6] = m[2]
+            else:
+                cpe_split[6] = '-'
+            bb.debug(1, "component[split]: version slpit {} / {}".format(cpe_split[5], cpe_split[6]))
```
JoergZeidler: Typo: `version slpit` → `version split`

classes/dependency-track.bbclass
HarryWaschkeit: damn ... fixed

classes/dependency-track.bbclass
HarryWaschkeit: would have been acceptable for me, but no objection against a dedicated check.
fixed.

classes/dependency-track.bbclass
HarryWaschkeit: fixed

classes/dependency-track.bbclass
HarryWaschkeit: fixed

classes/dependency-track.bbclass
HarryWaschkeit: fixed

classes/dependency-track.bbclass
HarryWaschkeit: fixed

classes/dependency-track.bbclass
mlilien: claude:
bb.error() logs at error level but does not raise / abort the task (only bb.fatal()
 raises BBHandledException). So when the regex fails to match (any version not starting with
 a digit), execution continues to cpe_split[5] = m[1] and dies with a raw
 'NoneType' object is not subscriptable traceback — the opposite of a clean, handled error.

 Reproduced: re.search(pat, "v1.2.3") → None; next line raises TypeError.

 For the currently-configured components (sudo, openssh) versions are numeric, so this
 won't fire today — it's a latent trap for the next component someone enables splitting on.

 Recommended fix (in classes/dependency-track.bbclass)

 Pick one:

 - Fail loudly (matches author's evident intent, smallest change):
 if m is None:
     bb.fatal("component {}: version {} cannot be split for CPE".format(name, full_ver))
 - bb.fatal logs and aborts cleanly instead of a traceback two lines later.
 - Or fail soft (more resilient — a single odd version doesn't break the whole SBOM step):
 if m is None:
     bb.warn("component {}: version {} not splittable; leaving CPE unsplit".format(name, full_ver))
     cpe_split[6] = '*'
 else:
     cpe_split[5] = m[1]
     cpe_split[6] = m[2] if m.groups()[1] is not None else '-'
     bb.debug(1, "component[split]: version split {} / {}".format(cpe_split[5], cpe_split[6]))

 Recommendation: fail-soft fallback — losing precise matching for one unexpected component is
 better than failing the entire image's SBOM collection.

mlilien (commented):
somehow the last review was not treated as one, but as a comment only
https://github.com/omnect/meta-omnect/pull/657#discussion_r3418952076

classes/dependency-track.bbclass
HarryWaschkeit: changed to failing loudly because we definitely don't want to have an sbom that is incorrect and moreover track version formats to see early when our sbom generation should still have issues.

mlilien (approved):

JoergZeidler (approved):

PR #667, opened 2026-07-16
chore(adu): update iot-hub-device-update to 1.4.0

JanZachmann:
## Summary
- Bump `iot-hub-device-update` recipe and the `omnect_*.patch` from 1.2.6 to 1.4.0.
- Drop the `adu-swupdate.sh` removal step in `do_install`; upstream deleted the script because the `microsoft/swupdate:1` handler is deprecated.
- Carry omnect fixes for cancel and cancel-to-replace of consent-gated updates (issue #511), replacing the old SIGUSR1 agent-restart workaround:
  - consent step returns `Failure_Cancelled` on cancel instead of restarting the agent;
  - `workflow_transfer_data` rebuilds step children on replacement so the new deployment uses its own `installedCriteria` instead of the replaced one's.

### DIFF
--- a/recipes-azure-iot/iot-hub-device-update/iot-hub-device-update_1.4.0.bb
+++ b/recipes-azure-iot/iot-hub-device-update/iot-hub-device-update_1.4.0.bb
@@ -14,7 +14,7 @@ def aduc_version(d, idx):
     return parts[idx]
 
 SRC_URI = " \
-  git://github.com/azure/iot-hub-device-update.git;protocol=https;tag=1.2.6;nobranch=1 \
+  git://github.com/azure/iot-hub-device-update.git;protocol=https;tag=1.4.0;nobranch=1 \
   file://deviceupdate-agent.service \
   file://deviceupdate-agent.timer \
   file://du-config.json \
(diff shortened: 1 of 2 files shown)


JoergZeidler (approved):

recipes-azure-iot/iot-hub-device-update/iot-hub-device-update_1.4.0.bb
```diff
@@ -119,9 +119,6 @@ do_install:append() {
   install -d ${D}${sysconfdir}/omnect/consent/swupdate
   install -m 0660 -o adu -g adu ${S}/src/extensions/step_handlers/swupdate_consent_handler/files/user_consent.json ${D}${sysconfdir}/omnect/consent/swupdate/
 
-  # delete adu-swupdate.sh
-  rm ${D}${bindir}/adu-swupdate.sh
```
mlilien, line 123: ~~is this not there anymore?~~ never mind, it's documented in the PR description. what about our handlers though?

JoergZeidler: they never used this script...

mlilien: aren't they an adapted copy of that script?

JoergZeidler: ok, we used it in older swupdate-handler v1, with v2 we use only our own script

mlilien (approved):

PR #668, opened 2026-07-17
feat: introduced more image wide hardening (mainly RELRO)

HarryWaschkeit:
Hardening aspect RELRO protects attacks that modify addresses of dynamically linked functions which are stored in the global offset table (GOT) by making these tables read-only after relocation.

This commit introduces image wide RELRO settings (where applicable) with two exceptions:

- go compiler for 32 bit ARM architectures doesn't support position independent code (PCIE) generation and hence no complete RELRO support exists; this affects applications runc as well as containerd-shim-runc-v2
- for other architectures (x86-64 and arm64) runc has ful RELRO support, but containerd-shim-runc-v2 is by design deliberately excepted from that via its recipe in meta-virtualization

### DIFF
--- a/kas/distro/omnect-os.yaml
+++ b/kas/distro/omnect-os.yaml
@@ -100,5 +100,8 @@ env:
 
 local_conf_header:
   meta-omnect_distro_omnect-os: |
+    # take advantage of openembedded-core's hardening settings
+    require conf/distro/include/security_flags.inc
+    
     # For more options have a look into "poky/meta-poky/conf/local.conf.sample" resp.
     # "openebedded-core/meta/conf/local.conf.sample".
--- a/kas/distro/omnect-os.yaml
+++ b/kas/distro/omnect-os.yaml
@@ -102,6 +102,6 @@ local_conf_header:
   meta-omnect_distro_omnect-os: |
     # take advantage of openembedded-core's hardening settings
     require conf/distro/include/security_flags.inc
-    
+
     # For more options have a look into "poky/meta-poky/conf/local.conf.sample" resp.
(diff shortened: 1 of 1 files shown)


kas/distro/omnect-os.yaml
```diff
@@ -100,5 +100,8 @@ env:
 
 local_conf_header:
   meta-omnect_distro_omnect-os: |
+    # take advantage of openembedded-core's hardening settings
+    require conf/distro/include/security_flags.inc
```
mlilien: file not commited

kas/distro/omnect-os.yaml
```diff
@@ -100,5 +100,8 @@ env:
...
+
```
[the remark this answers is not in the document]
HarryWaschkeit: fixed.

kas/distro/omnect-os.yaml
HarryWaschkeit: as the comment says,:this file is part of openembedded-core, so nothing to commit here.

mlilien: it doesn't say that, but i got it now

mlilien (approved):

kas/distro/omnect-os.yaml
HarryWaschkeit: I thought the comment was clear; if you have a suggestion to improve it I'll happily change it.

mlilien: it's not about the comment, it's more about the require path ... you could make it explicit

HarryWaschkeit: ok, now I understand, changed,

mlilien (approved):

PR #674, opened 2026-07-23
feat: wrynose integration

mlilien:


dynamic-layers/raspberrypi/recipes-bsp/u-boot/u-boot/enable-reset-info-cmd-fragment.cfg
```diff
@@ -1,2 +1,2 @@
-CONFIG_DISPLAY_CPUINFO=y
+# CONFIG_DISPLAY_CPUINFO is not set
```
mlilien, line 1: on purpose, rpi hangs when activating it

recipes-omnect/omnect-device-service/omnect-device-service_0.44.1.bb
```diff
@@ -6,11 +6,13 @@ inherit cargo
 # DEFAULT_PREFERENCE = "-1"
 
 # how to get omnect-device-service could be as easy as but default to a git checkout:
-# SRC_URI += "crate://crates.io/omnect-device-service/0.44.0"
-SRC_URI += "git://github.com/omnect/omnect-device-service.git;protocol=https;nobranch=1;branch=main"
-SRCREV = "4d6be836f1b47d55e7479eae28d251b6a64348f2"
+# SRC_URI += "crate://crates.io/omnect-device-service/0.44.1"
+SRC_URI += "git://github.com/mlilien/omnect-device-service.git;protocol=https;nobranch=1;branch=main"
```
JanZachmann: **critical:** SRC_URI points to the personal fork. SRCREV `2a83cef` is exactly the `omnect/omnect-device-service` 0.44.1 tag commit, so only the URL needs to go back to `github.com/omnect/omnect-device-service.git`.

recipes-omnect/initrdscripts/omnect-os-initramfs/fs-mount
```diff
@@ -47,7 +47,12 @@ run_fs_mount()
     setup_data_partition "${ROOTFS_DIR}" "${data_opts}" || return 1
 
     # mount tmpfs
-    run_cmd mount -t tmpfs -o mode=0755,nodev,nosuid,strictatime tmpfs  ${ROOTFS_DIR}/run || return 1
+    # /run is mounted by the initramfs-framework and relocated onto the rootfs
+    # (mount --move /run ${ROOTFS_DIR}/run) in its finish module before switch_root.
+    # Mounting our own tmpfs on ${ROOTFS_DIR}/run here would be shadowed by that
+    # move (losing anything written to it), so we only re-apply our hardened
+    # options to the framework's /run instead.
+    run_cmd mount -o remount,mode=0755,nodev,nosuid,strictatime /run
```
JanZachmann, line 55: **critical:** the old mount line had `|| return 1`; the remount does not. If the remount fails, boot continues with `/run` missing `nodev,nosuid,strictatime`. If `/run` was never mounted at all, everything written there later (e.g. `omnect-os-initramfs.json`) vanishes at `switch_root` without any error. Please restore `|| return 1`.

recipes-core/base-files/base-files_%.bbappend
```diff
@@ -2,6 +2,6 @@ FILESEXTRAPATHS:prepend := "${LAYERDIR_omnect}/files/device_caps:"
 SRC_URI += "file://${MACHINE}.json"
 do_install:append() {
     sed -i 's/^export PS1=\(.*\)$/[ -z \"\$PS1\" ] \&\& export PS1=\1/' ${D}/etc/skel/.bashrc
-    install -m 0644 -D ${WORKDIR}/${MACHINE}.json ${D}${sysconfdir}/omnect/device_caps.json
+    install -m 0644 -D ${UNPACKDIR}/${MACHINE}.json ${D}${sysconfdir}/omnect/device_caps.json
 }
 FILES_${PN} += "${sysconfdir}/omnect/device_caps.json"
```
JanZachmann, line 7: `FILES_${PN}` is the pre-kirkstone override syntax — this line is ignored on wrynose. It works anyway because oe-core base-files packages `/`, but it should be `FILES:${PN} +=` (or be deleted).

conf/machine/genericx86-64.extra.conf
```diff
@@ -71,6 +70,14 @@ MACHINE_EXTRA_RRECOMMENDS:remove = "linux-firmware"
 
 DISTRO_FEATURES:append = " efi-secure-boot modsign"
 
+# We verify the grub config with SELoader (PKCS7 .p7b), matching OMNECT_GRUB_EFI_SB_FILES
+# and the LockDown/SELoader artifacts below. wrynose's meta-secure-core (commit 6d83fbf)
```
JanZachmann, line 74: Commit hash and scarthgap-vs-wrynose history in a code comment. Keep only the mechanism, e.g.: "meta-secure-core defaults to grub's built-in GPG .sig check; we verify the grub config with SELoader (PKCS7 .p7b) instead, so pin both switches."

recipes-bsp/bootloader-versioned/bootloader-versioned.bb
```diff
@@ -1,7 +1,10 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
+# file-only recipe: sources land in ${UNPACKDIR}, not the default ${UNPACKDIR}/${BP}.
```
JanZachmann, line 3: "not the default ${UNPACKDIR}/${BP}" describes what it is not. "file-only recipe: sources are unpacked directly into ${UNPACKDIR}." is enough. Same comment in 7 more recipes (bootloader-env, grub-env, omnect-os-initramfs-scripts, omnect-base-files, omnect-first-boot, omnect-wifi-commissioning, tpm-udev).

recipes-omnect/images/omnect-os-initramfs.bb
```diff
@@ -13,6 +13,12 @@ inherit omnect_initramfs
 
 IMAGE_NAME = "${OMNECT_INITRAMFS_IMAGE_NAME}"
 
+# This image is a cpio unpacked into RAM, not a read-write filesystem, so the default
+# 1.3 free-space overhead is meaningless here - it only inflated the INITRAMFS_MAXSIZE
+# check (the wrynose component growth then tipped it over 128M). Size the check to the
```
JanZachmann, line 18: The 1.3 / 128M values check out, but the parenthetical is transient history ("the wrynose growth then tipped it over"). Suggest: "The image is a cpio unpacked into RAM; the default 1.3 free-space overhead only inflates the INITRAMFS_MAXSIZE check. Size the check to the real content; INITRAMFS_MAXSIZE stays as the RAM guard."

recipes-devtools/bindgen/bindgen_0.72.1.bb
```diff
@@ -0,0 +1,122 @@
+# Auto-Generated by cargo-bitbake 0.3.16-alpha.0
+#
+# Bumped 0.60.1 -> 0.72.1 for the wrynose migration: iot-identity-service main is
```
JanZachmann, line 3: "Bumped 0.60.1 -> 0.72.1 for the wrynose migration" is change narrative; start at the requirement: "iot-identity-service main is Rust edition 2024, where extern blocks must be written 'unsafe extern \"C\"'; older bindgen emits the old form and the generated code fails to compile. 0.72.1 is the version aziot's ci/install-build-deps.sh pins." Same pattern in cbindgen_0.29.2.bb.

recipes-azure-iot/azure-identityd/aziot-identityd_git.bb
```diff
@@ -0,0 +1,676 @@
+# Pinned to iot-identity-service for the wrynose migration: SRCREV is the upstream
+# 1.6.0-rc.1 tag commit (the tip of main at pin time).
+#
+# The 1.5.6 crate set pinned bindgen 0.69.4 / clang-sys 1.6.1; clang-sys 1.6.1
+# predates oe-core's clang and cannot walk its libclang AST, so the tss2 type
+# bindings (tpm/tss-minimal/types-sys) came out opaque (every TPM2 struct as a
+# field-less `_address` blob) and aziot-identityd failed to compile. main pins
+# bindgen 0.72.1 / clang-sys 1.8.1, which parse the same headers correctly.
+# Crate list below regenerated from main's Cargo.lock; re-pin to a release tag
+# once one ships that carries these crate versions.
```
JanZachmann, line 10: Transitional pin — our convention is the explicit marker: `# TODO: remove me, as soon as an iot-identity-service release tag with these crate versions is CR: re-pin SRCREV to that tag.` The 1.5.6 history above can be shortened.

recipes-azure-iot/azure-identityd/aziot-identityd_git.bb
```diff
@@ -0,0 +1,676 @@
...
+inherit cargo
+
+# ~rc.1 sorts before a future 1.6.0 release; +git marks the git/SRCREV pin.
+PV = "1.6.0~rc.1+git"
+
+SRC_URI += "gitsm://github.com/Azure/iot-identity-service.git;protocol=https;nobranch=1"
+SRCREV = "1a7a6e70c3a20389fd24090bba201b7fdd7de582"
```
JanZachmann, line 17: Upstream has tagged 1.6.0 by now (`a6b1d16`). Worth re-pinning to the release and going back to a versioned recipe — here or as a follow-up.

conf/machine/raspberrypi4-64.extra.conf
```diff
@@ -12,4 +12,4 @@ UBOOT_ARCH = "arm64"
 
 # no differentiation for now between release an devel images concerning rootfs
 # size
-OMNECT_PART_SIZE_ROOTFS ?= "${@bb.utils.contains('OMNECT_RELEASE_IMAGE', '1', '760832', '760832', d)}"
+OMNECT_PART_SIZE_ROOTFS ?= "${@bb.utils.contains('OMNECT_RELEASE_IMAGE', '1', '760832', '900000', d)}"
```
JanZachmann, line 15: The comment says no differentiation, but the line now differentiates (760832 vs 900000). Same in phygate-tauri-l-imx8mm-2.extra.conf and genericx86-64.extra.conf. Suggest deleting the comment or: "devel images get a larger rootfs partition". (Also "release an devel" typo.)

conf/distro/include/omnect-os-fixes.conf
```diff
@@ -8,3 +8,22 @@ DEPENDS:append:pn-libx11 = " autoconf-archive-native"
 # lshw: fix lshw e.g. on tauri-l
 FILESEXTRAPATHS:prepend:pn-lshw := "${LAYERDIR_omnect}/files/patches:"
 SRC_URI:append:pn-lshw = " file://lshw_github_PR87.patch"
+
+# wrynose promoted the patch-status QA check to a fatal error. Our local recipe
+# patches carry no Upstream-Status; silence the check (remove from both ERROR_QA and
+# WARN_QA) rather than annotate every patch.
+ERROR_QA:remove = "patch-status"
+WARN_QA:remove = "patch-status"
```
JanZachmann, line 16: This silences patch-status for all layers. The per-patch `Upstream-Status:` annotation is the maintained convention — the new wic patch in this PR already carries it. Policy call, just flagging it.

JanZachmann (commented):
Review focus: idiomatic yocto/bitbake, comment/language rules, scarthgap->wrynose (5.0.18 -> 6.0.2) migration completeness, inconsistencies.

**Overall:** the migration is thorough and internally consistent. Kas pins verified against upstream (oe-core/bitbake peel exactly from tag yocto-6.0.2, no layer left on an old branch); every deleted workaround maps to an upstream wrynose fix or the dropped rust 1.93 backport; WORKDIR->UNPACKDIR is complete (no `${WORKDIR}` refs left); all component SRCREVs match their claimed tags. Two blockers, both inline: the fork URL in omnect-device-service and the missing `|| return 1` in fs-mount.

**Should still be changed in this PR (lines outside the diff, so no inline comment possible):**

- README.md:102,108 — still says "compatible with ... `scarthgap`" and "versioned `5.0.x.y`". This PR bumps OE_VERSION to 6.0.2, so these lines need updating to wrynose / `6.0.x.y`.
- conf/machine/include/phytec-imx8mm.inc:62-72 — this PR drops meta-imx-sdk from kas, which turns the `meta-imx/meta-imx-sdk/...` BBMASK entries into no-ops (BBMASK on paths outside BBLAYERS does nothing). The stale entries can go.
- recipes-devtools/toml-cli/toml-cli-0.2.3.inc:1 and cargo-cyclonedx-0.5.7.inc:1 — "not compatible with scarthgap per se" should say wrynose or drop the release name.

**Noticed while reviewing, but the code is already on main (follow-up material, not this PR):**

- initrdscripts resize-data — three real issues: (1) lines 30-36: if a boot is interrupted between `parted resizepart` and `resize2fs`, the next boot sees no free sectors, sets the flag, and the filesystem stays small forever — run `resize2fs` (idempotent) in that branch too; (2) line 56: `[ $? -gt 1 ]` treats e2fsck exit 2 ("errors corrected, reboot needed") as fatal — real failures start at 4, use `-ge 4`; (3) line 42: `sgdisk -e` is the only command in the block without `|| return 1`.
- systemd_%.bbappend:65 — the coredump `Storage=none` sed silently no-ops if the conf template format drifts; suggest the `grep -q ... || bbfatal` guard this repo already uses in wifi-commissioning-service.inc.
- omnect-os-distro.conf:31-32 — bitbake 2.18 supports `addpylib`; a `lib/omnect/device_cap()` helper could call `bb.parse.mark_dependency()` and parse the JSON once instead of on every `DISTRO_FEATURES` expansion.
- doc/mac_lsm.md:25-26 — says the userspace install is gated on the `apparmor` DISTRO_FEATURE, but `OMNECT_MAC_USERSPACE` in omnect-os-image.bb is unconditional. Also line 6: "out of the box" -> "by default"; README.md:501: "levers" -> "settings".
- omnect-wifi-commissioning-start.sh:30 — `True`/`TRUE`/CRLF values of `WCS_DISABLE_BLE` silently keep BLE enabled; log the unmatched value.
- omnect-os-update-image.bb:58 — the `grubenv` skip has no why-comment.
- files/wic/omnect-os.rpi.wks.inc:5 — bare `include` while the `.wks.in` files use `${LAYERDIR_omnect}/...`; works, but the asymmetry invites a wrong copy later.

Cross-repo note: the omnect-os CI whitelists are organized per release (dunfell/kirkstone/scarthgap) — a `wrynose/` set will be needed there.

conf/distro/include/omnect-os-fixes.conf
```diff
@@ -8,3 +8,22 @@ DEPENDS:append:pn-libx11 = " autoconf-archive-native"
...
+
+# wrynose's oe-core also makes the buildpaths check fatal. Some Rust recipes (e.g. the
+# aziot cc-crate openssl engine) leak a ${WORKDIR} path into -dbg debug info that
+# escapes the debug-prefix-map remap; downgrade to a warning rather than chase the
+# per-crate remap. The runtime (non-dbg) binaries are clean.
```
JanZachmann, line 21: "The runtime (non-dbg) binaries are clean." is reassurance, not mechanism — can be cut. Alternative to the global downgrade: scope it with `INSANE_SKIP:pn-<recipe>` so the check stays fatal for everything else.

recipes-bsp/bootloader-versioned/bootloader-versioned.grub.inc
```diff
@@ -16,12 +16,13 @@ OMNECT_BOOTLOADER_CHECKSUM_FILES += "${LAYERDIR_efi-secure-boot}/recipes-bsp/shi
 inherit deploy omnect_bootloader_versioning
 
 do_install() {
-    install -m 0755 -D ${WORKDIR}/omnect_get_bootloader_version.grub.sh ${D}/usr/bin/omnect_get_bootloader_version.sh
+    install -m 0755 -D ${UNPACKDIR}/omnect_get_bootloader_version.grub.sh ${D}/usr/bin/omnect_get_bootloader_version.sh
```
JanZachmann, line 19: `${D}/usr/bin/` → `${D}${bindir}/`.

dynamic-layers/phytec/recipes-bsp/u-boot/u-boot-phytec-imx_%.bbappend
```diff
@@ -32,7 +32,47 @@ OMNECT_BOOTLOADER_CHECKSUM_FILES = "${OMNECT_BOOTLOADER_RECIPE_PATH}"
 
 inherit omnect_uboot_configure_env omnect_bootloader
 
+# wrynose's SWIG 4.3 made SWIG_Python_AppendOutput take a 3rd arg; U-Boot 2024.04's
+# bundled dtc/pylibfdt predates that, so its pylibfdt build fails to compile. Switch
```
JanZachmann, line 36: The guarded seds here silently no-op when upstream drifts. Both fixes exist upstream; backport patches with `Upstream-Status: Backport` would fail loudly instead of shipping a half-fixed build.

conf/distro/include/omnect-os-required.conf
```diff
@@ -1 +1 @@
-REQUIRED_VERSION_openssl ?= "3.5.6"
+REQUIRED_VERSION_openssl ?= "3.5.7"
```
JanZachmann, line 1: The exact pin fails the build if a future 6.0.x point release micro-bumps openssl. If the intent is "stay on the 3.5 LTS series", `3.5.%` survives point releases. A short why-comment would also help here.

dynamic-layers/raspberrypi/recipes-connectivity/pi-bluetooth/pi-bluetooth_%.bbappend
```diff
@@ -1,4 +1,3 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
```
JanZachmann, line 1: The dropped patch file `0001-Fix-device-dependency-in-hciuart.service.patch` is still in the tree — delete it.

dynamic-layers/raspberrypi/recipes-bsp/u-boot/u-boot/add-reset-info.patch
```diff
@@ -1,19 +1,22 @@
 diff --git a/arch/arm/mach-bcm283x/Makefile b/arch/arm/mach-bcm283x/Makefile
```
JanZachmann, line 1: Leftovers: `cmd/rstinfo.c` keeps an unused `extern int print_cpuinfo(void);`, and the `init.c` hunk prints inside `print_cpuinfo()`, which is compiled out by the cfg fragment (`# CONFIG_DISPLAY_CPUINFO is not set`). Both can go.

classes/omnect_uboot_configure_env.bbclass
```diff
@@ -28,11 +28,14 @@ do_compile:append() {
         mkenvimage="mkenvimage -r"
     fi
     if [ -n "${UBOOT_CONFIG}" ]; then
-        # this is based on the handling in openembedded-core/meta/recipes-bsp/u-boot/u-boot.inc
+        # openembedded-core's u-boot.inc now builds into O=${B}/${config}-${type},
+        # so the initial env lives in that -${type}-suffixed dir, not ${config}/.
+        # UBOOT_MACHINE/UBOOT_CONFIG only ever hold a single config here.
```
JanZachmann, line 33: Since only the first word of each variable is wanted, the nested loops with `break 2` can be:
```sh
set -- ${UBOOT_MACHINE}; config=$1
set -- ${UBOOT_CONFIG}; type=$1
```

classes/omnect_rust_azure-iot-sdk_deps.bbclass
```diff
@@ -1,6 +1,17 @@
+# wrynose dropped the custom rust-llvm (which carried clang); bindgen now gets
```
JanZachmann, line 1: Misattribution: the custom rust-llvm was our own recipe (deleted in this PR), not something the wrynose release dropped. Suggest: "the custom rust-llvm (which carried clang) was dropped with the wrynose upgrade; bindgen now gets libclang from oe-core's clang-native."

conf/machine/genericx86-64.extra.conf
```diff
@@ -71,6 +70,14 @@ MACHINE_EXTRA_RRECOMMENDS:remove = "linux-firmware"
...
+# flipped the defaults to UEFI_SELOADER=0 + GRUB_SIGN_VERIFY=1 (grub's built-in GPG .sig
+# check); scarthgap had UEFI_SELOADER=1 + GRUB_SIGN_VERIFY=0. The two are mutually
+# exclusive, so pin both to restore the SELoader path.
+UEFI_SELOADER = "1"
+GRUB_SIGN_VERIFY = "0"
+
 # for patched wic plugin bootimg-efi.py:
```
JanZachmann, line 81: Comment drift: the patched plugin now lives in the standalone wic recipe (`src/wic/plugins/source/bootimg_efi.py`), patched via `recipes-support/wic/wic_%.bbappend`.

conf/distro/include/omnect-os-distro.conf
```diff
@@ -179,6 +178,9 @@ PACKAGECONFIG:remove:pn-systemd = "logind pam"
 # gobject-introspection dependency in avahi + polkit
 CODEGEN_PYTHON_RDEPENDS:pn-glib-2.0 = ""
 RDEPENDS:${PN}:pn-gobject-introspection = ""
+# Emptying RDEPENDS also drops the auto libglib shlib dep; wrynose made file-rdeps
+# fatal. glib is always in the image (avahi/polkit depend on it directly), so skip it.
+INSANE_SKIP:${PN}:pn-gobject-introspection = "file-rdeps"
 PACKAGECONFIG:remove:pn-apparmor = "python"
 
 # disable glib-2.0 dependency to shared-mime-info
```
JanZachmann, line 186: This `pn-apparmor` line sits in the glib block; the other apparmor tweaks (perl removal, PERLCONFIGTARGET) are ~20 lines up — move it next to them.

recipes-omnect/omnect-device-service/omnect-device-service_0.44.1.bb
JanZachmann: i keep that one as a reminder

recipes-devtools/bindgen/bindgen_0.72.1.bb
mlilien: claude generated the recipe with the cargo-bitbake header line 1 ;) ... cargo-bitbake wouldn't add a comment in line 3.

recipes-azure-iot/azure-identityd/aziot-identityd_git.bb
mlilien: not as part of this PR. this should be done in an extra iotedge 1.6.0 update.

conf/distro/include/omnect-os-fixes.conf
mlilien: won't change.

classes/omnect_rust_azure-iot-sdk_deps.bbclass
mlilien: rm comment

recipes-omnect/omnect-device-service/omnect-device-service_0.44.1.bb
mlilien: done

PR #677, opened 2026-07-27
fix(aziot-identityd): let precondition service reach failed state

JanZachmann:
## Summary
- add StartLimitBurst=5 / StartLimitIntervalSec=120 to aziot-identityd-precondition.service so a persistent failure reaches 'failed' (~30 s) instead of looping in 'activating' forever
- add aziot-identityd-precondition.timer (OnUnitInactiveSec=10min) so transient failures still recover: the retry runs after the rate-limit window has passed

## Reason
With Restart=on-failure/RestartSec=5 and the default start limit (5 in 10 s) the rate limit could never be reached: a persistently failing precondition looped in 'activating' forever and never showed up in 'systemctl is-system-running'.

### DIFF
--- a/recipes-azure-iot/azure-identityd/aziot-identityd/aziot-identityd-precondition.service
+++ b/recipes-azure-iot/azure-identityd/aziot-identityd/aziot-identityd-precondition.service
@@ -17,10 +17,11 @@ ConditionPathExists=|/run/omnect-device-service/omnect_bootloader_updated
 ConditionFileNotEmpty=|!/etc/aziot/identityd/config.d/00-super.toml
 
 # with RestartSec=5 the default limit (5 starts per 10s) is never reached, so
-# the unit would loop in auto-restart forever instead of failing; keep the
-# window short, the aziot stack is ordered after this unit
+# the unit would loop in auto-restart forever instead of failing; the window
+# has to hold the whole burst, i.e. StartLimitBurst * (RestartSec + runtime of
+# one 'config apply')
 StartLimitBurst=5
-StartLimitIntervalSec=60
+StartLimitIntervalSec=120
 
 [Service]
 Type=oneshot
--- a/recipes-azure-iot/azure-identityd/aziot-identityd/aziot-identityd-precondition.service
+++ b/recipes-azure-iot/azure-identityd/aziot-identityd/aziot-identityd-precondition.service
@@ -15,6 +15,8 @@ ConditionPathExists=|/run/omnect-device-service/omnect_validate_update_failed
 ConditionPathExists=|/run/omnect-device-service/omnect_bootloader_updated
 
 ConditionFileNotEmpty=|!/etc/aziot/identityd/config.d/00-super.toml
+StartLimitBurst=5
+StartLimitIntervalSec=60
 
 [Service]
 Type=oneshot
--- a/recipes-azure-iot/azure-identityd/aziot-identityd/aziot-identityd-precondition.service
+++ b/recipes-azure-iot/azure-identityd/aziot-identityd/aziot-identityd-precondition.service
@@ -15,6 +15,10 @@ ConditionPathExists=|/run/omnect-device-service/omnect_validate_update_failed
 ConditionPathExists=|/run/omnect-device-service/omnect_bootloader_updated
 
 ConditionFileNotEmpty=|!/etc/aziot/identityd/config.d/00-super.toml
+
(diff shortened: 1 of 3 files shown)


recipes-azure-iot/azure-identityd/aziot-identityd/aziot-identityd-precondition.service
```diff
@@ -15,6 +15,8 @@ ConditionPathExists=|/run/omnect-device-service/omnect_validate_update_failed
 ConditionPathExists=|/run/omnect-device-service/omnect_bootloader_updated
 
 ConditionFileNotEmpty=|!/etc/aziot/identityd/config.d/00-super.toml
+StartLimitBurst=5
```
JoergZeidler, line 23: configuration differs to device-update-agent or omnect-device-service. why do you have used these settings?

JanZachmann: The two other units are long-running `Type=notify` daemons with `Restart=always` (`deviceupdate-agent` 10/120 RestartSec=5, `omnect-device-service` 10/120 RestartSec=10). A restart there is cheap and nothing is ordered after them, so a generous burst fits. This one is a `Type=oneshot` unit ordered `Before=` the whole aziot stack, which changes both requirements.

First, an override is needed at all: the layer sets no `DefaultStartLimit*`, so systemd defaults apply (5 starts per 10s). With `RestartSec=5` a cycle takes ~6-7s, so at most 2 starts fit in any 10s window - the limit is never reached, the unit loops in `auto-restart` forever and the system never becomes degraded. That is the bug this PR fixes.

Second, why the window is shorter than 10/120: while the unit sits in `auto-restart`, its `Before=` ordering holds the aziot services back. With 5/60 it reaches `failed` after ~35s and a retry succeeds ~30s later - both measured on the DUT in a smoke-test run, which also confirms `aziot config apply` fails in ~1-2s per attempt. With 10/120 it would take ~70s to fail and no retry could succeed before 120s. That matters inside the update-validation window (omnect-device-service default 300s, CI injects 90s), where authentication has to complete before the deadline, and it would roughly double the deadlines in `precondition_recovery_test.sh` (currently 90s/120s).

The `OnUnitInactiveSec=10min` retry timer is aligned with `deviceupdate-agent.timer`.

Added the reasoning to the unit file in c1186ef, since it was not visible from the values. If you prefer consistency with the daemons over failing fast, 10/120 reaches the limit as well - it only costs the extra boot delay and the doubled test deadlines, which I would raise in the same round.

JoergZeidler (approved):

HarryWaschkeit (approved):

recipes-azure-iot/azure-identityd/aziot-identityd/aziot-identityd-precondition.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Retry Azure IoT Identity Service Precondition
+
+[Timer]
+# must exceed the unit's start-limit window, else the retry is rejected as well
+OnUnitInactiveSec=10min
```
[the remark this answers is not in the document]
JanZachmann: `OnUnitInactiveSec=` does arm when the trigger unit is in `failed`. Checked against the systemd sources of the version in our image (255.21):

- `timer.c:462`: for `TIMER_UNIT_INACTIVE` the time base is `trigger->inactive_enter_timestamp`.
- `unit.c:2747-2748`: that timestamp is set on every transition into `UNIT_IS_INACTIVE_OR_FAILED`, and `unit.h:52` defines it as `IN_SET(t, UNIT_INACTIVE, UNIT_FAILED)`. So `failed` updates it just like `inactive`.

The start limit is also not sticky: `ratelimit_below()` (`basic/ratelimit.c:21`) opens a new window as soon as `now - begin > interval`, so a start attempt 10 min after a 120 s window is accepted without `reset-failed`.

Both were measured on the DUT in the smoke test for this change (omnect/omnect-os#816): it asserts the timer's `SubState` is `waiting` right after the unit reaches `failed`, and that a start succeeds after the window without `reset-failed`.

No steady-state retry loop either: with `RemainAfterExit=true` a successful triggered run leaves the unit `active (exited)`, and while the timer is in `running` it only re-arms when the trigger goes inactive-or-failed (`timer.c:797-802`). So the 10 min retry repeats only as long as the unit keeps failing.

The `OnFailure=` + `OnActiveSec=` variant would work as well, but it would replace a mechanism that is already verified on hardware, so keeping this one.

PR #678, opened 2026-07-31
fix(bootloader-env): block all fw_setenv options, not only -s/--script

JanZachmann:
## Summary
`bootloader_env.sh set` (u-boot variant) passes `--` to `fw_setenv` instead of comparing
key and value against a list of known flags. Same for `unset` and `get`, whose key reaches
`fw_setenv` / `fw_printenv` the same way.

Also updates omnect-device-service to 0.45.1, which carries the same fix for its own wrapper `sudo/fw_setenv_no_script.sh`. Recipe regenerated with cargo-bitbake; the crate set did not change.

## Reason
The blocklist covered `-s`, `--script`, `-s=*` and `--script=*`. `fw_setenv`
(libubootenv, `src/fw_printenv.c`) parses its arguments with `getopt_long` and the
optstring `Vc:f:s:nhm:`, which accepts attached values (`-sFILE`) and abbreviations
(`--scr=FILE`), so those forms passed the filter and reached `fw_setenv` as options.
GNU getopt also keeps scanning after the first positional argument, so the value
position is enough.

`recipes-azure-iot/iot-hub-device-update/iot-hub-device-update/adu-bootloader-env`
pins the key but passes the value unfiltered for `omnect_extra_bootargs` and
`omnect_validate_extra_bootargs`, so the `adu` user could write files as root — via a
script file (`-s`) or an attacker-chosen config (`-c`), the latter not covered by a
`-s`/`--script` blocklist at all.

This is the same defect that was fixed in omnect-device-service's `fw_setenv_no_script.sh`
(omnect/omnect-device-service#208); this script is the second copy of it. Both copies ship in the same image, so the ods pin belongs here — otherwise the image keeps the unfixed wrapper.

### DIFF
--- a/recipes-omnect/bootloader_env/bootloader-env/bootloader_env_u-boot.sh
+++ b/recipes-omnect/bootloader_env/bootloader-env/bootloader_env_u-boot.sh
@@ -11,7 +11,7 @@ function help() {
 function get() {
     [[ ${argsc} -ne 2 ]] && help && exit 1
     local key=${1}
-    local value=$(fw_printenv ${key})
+    local value=$(fw_printenv -- ${key})
     value=${value#${key}=}
     [[ -z "${value}" ]] && echo && exit 2
     echo ${value}
@@ -27,19 +27,17 @@ function set () {
     local key=${1}
     local value=${@:2}
 
-    for word in $key $value; do
-        if [[ "$word" == "-s" || "$word" == "--script" || "$word" == -s=* || "$word" == --script=* ]]; then
-            echo "Script-file mode is not allowed (flag: $word)"; exit 66;
-        fi
-    done
-
-    fw_setenv "${key}" "${value}"
+    # '--' ends option parsing, so key and value are always treated as data.
+    # this blocks script mode and every other option, e.g. an attacker-chosen
+    # config file, which a flag blocklist would miss (getopt accepts attached
+    # values like -sFILE and abbreviations like --scr=FILE)
+    fw_setenv -- "${key}" "${value}"
 }
 
 function unset() {
     [[ ${argsc} -ne 2 ]] && help && exit 1
     local key=${1}
-    fw_setenv "${key}"
+    fw_setenv -- "${key}"
 }
 
 [[ ${#} -lt 1 ]] && help && exit 1
--- a/recipes-omnect/bootloader_env/bootloader-env/bootloader_env_u-boot.sh
+++ b/recipes-omnect/bootloader_env/bootloader-env/bootloader_env_u-boot.sh
@@ -11,10 +11,10 @@ function help() {
 function get() {
     [[ ${argsc} -ne 2 ]] && help && exit 1
     local key=${1}
-    local value=$(fw_printenv -- ${key})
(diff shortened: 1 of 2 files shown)


recipes-omnect/bootloader_env/bootloader-env/bootloader_env_u-boot.sh
```diff
@@ -27,19 +27,17 @@ function set () {
...
-            echo "Script-file mode is not allowed (flag: $word)"; exit 66;
-        fi
-    done
-
-    fw_setenv "${key}" "${value}"
+    # '--' ends option parsing, so key and value are always treated as data.
+    # this blocks script mode and every other option, e.g. an attacker-chosen
+    # config file, which a flag blocklist would miss (getopt accepts attached
+    # values like -sFILE and abbreviations like --scr=FILE)
+    fw_setenv -- "${key}" "${value}"
```
HarryWaschkeit, line 34: just a note: using "--" as separator for options implies that the implementation uses a `getopt()` function that respects this.
while this is currently true (an very unlikely to change in the future), it still poses a small risk that this might change in future (maybe when transitioning to a different implementation for the binary).
at least the documentation for the tool does not explicitly mention this special treatment.
however, it would still be fine by me to run that risk, hoping that a different implementation would either handle `--` the same way or error on it if not supporting it.

recipes-omnect/bootloader_env/bootloader-env/bootloader_env_u-boot.sh
```diff
@@ -11,7 +11,7 @@ function help() {
 function get() {
     [[ ${argsc} -ne 2 ]] && help && exit 1
     local key=${1}
-    local value=$(fw_printenv ${key})
+    local value=$(fw_printenv -- ${key})
     value=${value#${key}=}
     [[ -z "${value}" ]] && echo && exit 2
     echo ${value}
```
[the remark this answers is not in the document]
JanZachmann: Applied in d90a2d5: quoted the key on `fw_printenv`, used `${value#"${key}"=}` for the prefix strip, and switched to `printf '%s\n'` for output.

The glob and word-split points are defense-in-depth here — the adu sudoers file pins the key to fixed literals (`omnect_extra_bootargs`, `omnect_validate_extra_bootargs`, …), so no attacker-chosen key reaches `get()` today. `printf` does fix a real read-back bug though: values with a leading `-n`/`-e`/`--` were truncated when `swupdate_handler_v2_u-boot.sh` reads them back through `get`.

recipes-omnect/bootloader_env/bootloader-env/bootloader_env_u-boot.sh
JanZachmann: Acknowledged. If libubootenv ever drops POSIX `--` handling we would see it in the smoke tests and revisit. The blocklist alternative I removed missed attached values (`-sFILE`) and abbreviations (`--scr=FILE`) entirely, so it was not the safer path either.

HarryWaschkeit (approved):

PR #683, opened 2026-08-10
fix(azure-identityd): wait for a working resolver before starting identityd

JanZachmann:
## Summary

`aziot-identityd.service` gets a second `ExecStartPre` which waits until every endpoint host configured in `/etc/aziot/config.toml` resolves - `cert_issuance.est.urls.default`, `provisioning.global_endpoint` and `provisioning.iothub_hostname`. A single lookup is capped at 5 s and the whole wait at 30 s. No endpoint configured means no wait, and on timeout the daemon starts anyway.

## Reason

identityd enrolls the `device-id` cert via EST and reaches out to DPS right at start. When the resolver cannot answer it gets `EAI_AGAIN`, exits 1 and is restarted; the second attempt succeeds. This hits the first boot after flashing or a factory reset - the state every CI DUT boots in. The failed start also fails the ADU agent health check and logs `[ERR!]` lines that are not in the smoke-test syslog whitelist, while neither the crash-loop check nor `systemctl is-system-running` can see it: a unit that failed once and recovered is `active` with `NRestarts=1`.

Seen twice on the rpi4 nightly, last in `test-image-job` 541. The lookup blocks for its full resolver timeout before it fails: `aziot-certd` starts at 18:09:14 and reports `Temporary failure in name resolution` at 18:09:34, identityd exits, and the restart at 18:09:40 succeeds. The rest of the network worked for the whole of those 20 s - the test runner opened dozens of ssh connections to the DUT between 18:09:15 and 18:09:34, and timesyncd had reached the lab time server at 18:09:08, before the window. So the queries were sent and stayed unanswered, which is a different thing from resolved having no server at all - that case fails immediately instead of blocking. Why the resolver was silent for those 20 s is not established; it is the lab resolver handed out by DHCP, and it answers in a few milliseconds when warm.

The unit already waits for `network-online.target`, and that target only says a link is configured. DHCP hands out the address of a DNS server, not a DNS server that answers, and no ordering can express "a name resolves" - which is why the wait has to attempt a lookup. It also means the gate does not depend on which cause it was: resolver not warm, resolved not ready or an upstream that is slow all end in the same wait.

Which endpoint identityd needs first depends on the provisioning method, and they are not the same domain, so every configured host is probed and each one still pending gets one try per pass - a host that never resolves must not starve the others. The single lookup is capped as well, because a failing lookup blocks for the resolver timeout: the unit sets no `TimeoutStartSec`, so its ceiling is the 90 s default, and running into it would produce the very failed start this gate exists to prevent.

The gate delays the start and never skips it. A condition-skipped activation would consume `aziot-identityd.socket`'s trigger limit of 20 within 2 s, and a failed socket is never activated again on its own (#680).

### DIFF
--- a/recipes-azure-iot/azure-identityd/aziot-identityd.inc
+++ b/recipes-azure-iot/azure-identityd/aziot-identityd.inc
@@ -20,6 +20,7 @@ SRC_URI += " \
     file://aziot-identityd-precondition.timer \
     file://iot-identity-service.conf \
     file://iot-identity-service-certd.template.toml \
+    file://omnect_wait_dns_ready.sh \
 "
 
 CARGO_BUILD_FLAGS += "--offline"
@@ -35,7 +36,9 @@ DEPENDS = " \
 "
 
 RDEPENDS:${PN} = " \
+    bash \
     libtss2-tcti-device \
+    toml-cli \
 "
 
 do_compile() {
@@ -84,6 +87,7 @@ do_install() {
     # binaries
     install -d -m 0755  ${D}${bindir}
     install -m 0755     ${B}/target/${CARGO_TARGET_SUBDIR}/aziotctl ${D}${bindir}
+    install -m 0755     ${WORKDIR}/omnect_wait_dns_ready.sh ${D}${bindir}
 
     install -d -m 0750 -g aziot ${D}${libexecdir}/aziot-identity-service
     install -m 0750 -g aziot ${B}/target/${CARGO_TARGET_SUBDIR}/aziotd ${D}${libexecdir}/aziot-identity-service
@@ -142,7 +146,7 @@ do_install() {
         -e 's#^After=\(.*\)$#After=\1\nConditionPathExists=/etc/aziot/config.toml\nConditionPathExists=/etc/aziot/identityd/config.d/00-super.toml#' \
         -e 's#@libexecdir@#/usr/libexec#g' \
         -e '/Environment=\(.*\)$/d' \
-	-e 's#^ExecStart=\(.*\)$#ExecStartPre=+-bash -c "/usr/sbin/omnect_service_log.sh start aziot-identityd"\nExecStart=\1\nExecStopPost=+-/bin/sh -c "/usr/sbin/omnect_service_log.sh stop aziot-identityd \\"$$SERVICE_RESULT\\" \\"$$EXIT_CODE\\" \\"$$EXIT_STATUS\\""#' \
+	-e 's#^ExecStart=\(.*\)$#ExecStartPre=+-bash -c "/usr/sbin/omnect_service_log.sh start aziot-identityd"\nExecStartPre=+-/usr/bin/omnect_wait_dns_ready.sh\nExecStart=\1\nExecStopPost=+-/bin/sh -c "/usr/sbin/omnect_service_log.sh stop aziot-identityd \\"$$SERVICE_RESULT\\" \\"$$EXIT_CODE\\" \\"$$EXIT_STATUS\\""#' \
         -e 's/^Requires=\(.*\)$/Requires=\1 network-online.target/' \
         ${D}${systemd_system_unitdir}/aziot-identityd.service
     install -m 0644     ${S}/identity/aziot-identityd/aziot-identityd.socket.in ${D}${systemd_system_unitdir}/aziot-identityd.socket
--- a/recipes-azure-iot/azure-identityd/aziot-identityd.inc
+++ b/recipes-azure-iot/azure-identityd/aziot-identityd.inc
@@ -147,7 +147,7 @@ do_install() {
         -e 's#^After=\(.*\)$#After=\1\nConditionPathExists=/etc/aziot/config.toml\nConditionPathExists=/etc/aziot/identityd/config.d/00-super.toml#' \
         -e 's#@libexecdir@#/usr/libexec#g' \
         -e '/Environment=\(.*\)$/d' \
-	-e 's#^ExecStart=\(.*\)$#ExecStartPre=+-bash -c "/usr/sbin/omnect_service_log.sh start aziot-identityd"\nExecStartPre=/usr/bin/omnect_wait_dns_ready.sh\nExecStart=\1\nExecStopPost=+-/bin/sh -c "/usr/sbin/omnect_service_log.sh stop aziot-identityd \\"$$SERVICE_RESULT\\" \\"$$EXIT_CODE\\" \\"$$EXIT_STATUS\\""#' \
+	-e 's#^ExecStart=\(.*\)$#ExecStartPre=+-bash -c "/usr/sbin/omnect_service_log.sh start aziot-identityd"\nExecStartPre=+/usr/bin/omnect_wait_dns_ready.sh\nExecStart=\1\nExecStopPost=+-/bin/sh -c "/usr/sbin/omnect_service_log.sh stop aziot-identityd \\"$$SERVICE_RESULT\\" \\"$$EXIT_CODE\\" \\"$$EXIT_STATUS\\""#' \
         -e 's/^Requires=\(.*\)$/Requires=\1 network-online.target/' \
         ${D}${systemd_system_unitdir}/aziot-identityd.service
     install -m 0644     ${S}/identity/aziot-identityd/aziot-identityd.socket.in ${D}${systemd_system_unitdir}/aziot-identityd.socket
--- a/recipes-azure-iot/azure-identityd/aziot-identityd.inc
+++ b/recipes-azure-iot/azure-identityd/aziot-identityd.inc
@@ -147,7 +147,7 @@ do_install() {
         -e 's#^After=\(.*\)$#After=\1\nConditionPathExists=/etc/aziot/config.toml\nConditionPathExists=/etc/aziot/identityd/config.d/00-super.toml#' \
         -e 's#@libexecdir@#/usr/libexec#g' \
         -e '/Environment=\(.*\)$/d' \
-	-e 's#^ExecStart=\(.*\)$#ExecStartPre=+-bash -c "/usr/sbin/omnect_service_log.sh start aziot-identityd"\nExecStartPre=+-/usr/bin/omnect_wait_dns_ready.sh\nExecStart=\1\nExecStopPost=+-/bin/sh -c "/usr/sbin/omnect_service_log.sh stop aziot-identityd \\"$$SERVICE_RESULT\\" \\"$$EXIT_CODE\\" \\"$$EXIT_STATUS\\""#' \
+	-e 's#^ExecStart=\(.*\)$#ExecStartPre=+-bash -c "/usr/sbin/omnect_service_log.sh start aziot-identityd"\nExecStartPre=/usr/bin/omnect_wait_dns_ready.sh\nExecStart=\1\nExecStopPost=+-/bin/sh -c "/usr/sbin/omnect_service_log.sh stop aziot-identityd \\"$$SERVICE_RESULT\\" \\"$$EXIT_CODE\\" \\"$$EXIT_STATUS\\""#' \
         -e 's/^Requires=\(.*\)$/Requires=\1 network-online.target/' \
         ${D}${systemd_system_unitdir}/aziot-identityd.service
     install -m 0644     ${S}/identity/aziot-identityd/aziot-identityd.socket.in ${D}${systemd_system_unitdir}/aziot-identityd.socket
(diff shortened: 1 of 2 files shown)


recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
```diff
@@ -0,0 +1,63 @@
...
+
+AZIOT_CONFIG=/etc/aziot/config.toml
+TIMEOUT_SECONDS=30
+LOOKUP_TIMEOUT_SECONDS=5
+POLL_INTERVAL_SECONDS=1
+
+# est and dps are urls, the iothub a bare host
+function url_host()
+{
+    local value="${1#*://}"
```
HarryWaschkeit: is there a reason to not use the "greedy" pattern matching as on all other pattern removals below?
to me this prefix removal is identical in usage as the (potential) user strip in line 17.

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
```diff
@@ -0,0 +1,63 @@
...
+    value="${value%%/*}"
+    value="${value##*@}"
+
+    echo "${value%%:*}"
+}
+
+# which endpoint is needed first depends on the provisioning method, and they
+# are not the same domain, so every configured one is probed
+declare -A seen=()
+pending=()
```
HarryWaschkeit: why seen and pending?
if I'm not mistaken, pending would be sufficient: it is the only variable operated on below the `for` loop

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
```diff
@@ -0,0 +1,63 @@
...
+    remaining=()
+
+    for host in "${pending[@]}"; do
+        timeout "${LOOKUP_TIMEOUT_SECONDS}" getent ahosts "${host}" > /dev/null \
+            || remaining+=("${host}")
+    done
+
+    pending=("${remaining[@]}")
+    [[ ${#pending[@]} -gt 0 ]] || break
+    sleep "${POLL_INTERVAL_SECONDS}"
```
HarryWaschkeit: what is the benefit to sleep here instead of directly looping again?
the timeout'ed getent should prevent too much system stress, so the script would probably simply finish faster.

recipes-azure-iot/azure-identityd/aziot-identityd.inc
```diff
@@ -142,7 +147,7 @@ do_install() {
         -e 's#^After=\(.*\)$#After=\1\nConditionPathExists=/etc/aziot/config.toml\nConditionPathExists=/etc/aziot/identityd/config.d/00-super.toml#' \
         -e 's#@libexecdir@#/usr/libexec#g' \
         -e '/Environment=\(.*\)$/d' \
-	-e 's#^ExecStart=\(.*\)$#ExecStartPre=+-bash -c "/usr/sbin/omnect_service_log.sh start aziot-identityd"\nExecStart=\1\nExecStopPost=+-/bin/sh -c "/usr/sbin/omnect_service_log.sh stop aziot-identityd \\"$$SERVICE_RESULT\\" \\"$$EXIT_CODE\\" \\"$$EXIT_STATUS\\""#' \
+	-e 's#^ExecStart=\(.*\)$#ExecStartPre=+-bash -c "/usr/sbin/omnect_service_log.sh start aziot-identityd"\nExecStartPre=+-/usr/bin/omnect_wait_dns_ready.sh\nExecStart=\1\nExecStopPost=+-/bin/sh -c "/usr/sbin/omnect_service_log.sh stop aziot-identityd \\"$$SERVICE_RESULT\\" \\"$$EXIT_CODE\\" \\"$$EXIT_STATUS\\""#' \
```
HarryWaschkeit: why do we suppress a fail exit status of `omnect_wait_dns_ready.sh`?
as far as I can see it will always be 0? (ends explicitly with `exit 0` only and doesn't terminate on failing commands)

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
JanZachmann: Different case, so a different operator. The first `://` separates the scheme, the last `@` separates the userinfo, so the scheme strip has to be non-greedy and the userinfo strip greedy.

```
https://host.example.com/redirect?target=http://evil.example.net/x
  ${u#*://}   -> host.example.com
  ${u##*://}  -> evil.example.net
```

The path strip runs after the scheme strip, so a second `://` in a query or path is still in the string at that point.

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
JanZachmann: Dropped in 303f5e3, and the dedup with it. The three endpoints are always different domains (edge CA, `global.azure-devices-provisioning.net`, `*.azure-devices.net`), so a host cannot appear twice, and if it ever did the only cost is one extra `getent` per pass.

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
JanZachmann: The `timeout` only bounds the slow case. The common failure is fast: `getent ahosts` on an unresolvable name returns in a few milliseconds.

```
$ time getent ahosts does-not-exist.invalid
real 0m0.004s   rc=2
```

Without the sleep the loop would fork `timeout` + `getent` a few hundred times a second for the full 30 s, during boot. Added a comment for it in 303f5e3.

The cost is at most one second of extra latency, and only on a pass where a host did not resolve.

recipes-azure-iot/azure-identityd/aziot-identityd.inc
JanZachmann: Correct that the script itself always exits 0 - it has no `set -e` and ends on `exit 0`. The `-` covers what the script cannot: systemd counts a failed exec as a failed `ExecStartPre`, so a missing, non-executable or killed script would fail the identityd start. That is the exact failure this PR removes, and the gate must never be able to cause it. It also matches the `omnect_service_log.sh` line above.

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
```diff
@@ -0,0 +1,63 @@
...
+           provisioning.global_endpoint \
+           provisioning.iothub_hostname; do
+    # toml prints nothing for an absent key
+    value=$(toml get -r "${AZIOT_CONFIG}" "${key}")
+    [[ -n "${value}" ]] || continue
+
+    host=$(url_host "${value}")
+    [[ -n "${host}" && -z "${seen[${host}]}" ]] || continue
+    seen["${host}"]=1
+    pending+=("${host}")
```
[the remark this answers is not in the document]
JanZachmann: This is wrong. toml-cli at the pinned `SRCREV=ea69e9d2ca4f0f858110dc7a5ae28bcb918c07fb` exits non-zero with no output for an absent key, and never prints `null`:

```rust
// walk_tpath
Name(n) => item = item.get(n)?,   // None as soon as a segment is missing
// get
let item = item.ok_or(SilentError::KeyNotFound { key: query.into() })?;
```

`SilentError` is documented there as "An error that should cause a failure exit, but no message on stderr".

`omnect_get_deviceid.sh` is not evidence of the opposite - its `== "null"` comparison is a bug for the same reason, and is being fixed in #684.

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
HarryWaschkeit: a second `://` is as (un)likely to me as a second `@` so I don't see any difference here, but so be it.

recipes-azure-iot/azure-identityd/aziot-identityd.inc
HarryWaschkeit: I don't buy it: this PR adds both, the script to the read-only rootfs as well as the ExecStartPre extension to the service file, a non-existing (executable) script would be a system error that cannot be handled at all imho.
so, why/how should this `-` matter here?

HarryWaschkeit (commented):
see comment https://github.com/omnect/meta-omnect/pull/683/changes#r3810905041

recipes-azure-iot/azure-identityd/aziot-identityd.inc
JanZachmann: You are right, dropped both prefixes in 2b31533.

`-` only ever covered a missing or non-executable script, which is a broken image and should fail loud rather than be swallowed. The script itself cannot return non-zero, so nothing else was being suppressed.

`+` is gone too, for the same "buys nothing" reason: `/etc/aziot` is `0775 root:aziot` and `config.toml` is `0664 root:aziot`, identityd reads that same file as `aziotid`, and the unit has no sandboxing directives - so `toml`, `timeout` and `getent` all work without privileges. Say if you would rather keep `+` as a guard against the perms changing.

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
JanZachmann: It is not about likelihood, it is which occurrence the grammar makes the separator. The scheme ends at the *first* `:` (RFC 3986 `scheme ":" hier-part`), so anything later is data. The userinfo ends at the *last* `@`, because `@` is not legal unencoded in a host, so a `@` further right cannot be part of one.

For our three endpoints neither case occurs and both operators would give the same answer today, so this is only about not being wrong if one ever shows up.

recipes-azure-iot/azure-identityd/aziot-identityd.inc
HarryWaschkeit: given that other service file changes do use `+` I would prefer leaving it here too for uniformity.

HarryWaschkeit (commented):
see https://github.com/omnect/meta-omnect/pull/683#discussion_r3811734069

recipes-azure-iot/azure-identityd/aziot-identityd.inc
JanZachmann: Restored in a9548a0, so the line is now `ExecStartPre=+/usr/bin/omnect_wait_dns_ready.sh` - `+` back for uniformity, `-` gone.

recipes-azure-iot/azure-identityd/aziot-identityd/omnect_wait_dns_ready.sh
JanZachmann: Correcting my own number here. I measured `getent ahosts does-not-exist.invalid`, which is rejected locally and fails in milliseconds, and generalised from it. The incident this PR is about is the opposite case: in `test-image-job` 541 the lookup blocked for the full resolver timeout - `aziot-certd` starts at 18:09:14 and the failure lands at 18:09:34. Under our 5 s cap a pass would be 5 s per host, not a spin.

So the "few hundred forks a second" argument only holds for the fast-fail case, which is real but is not the one in the log: systemd-resolved with no configured server returns `EAI_AGAIN` immediately. The sleep is there for that case. If you would rather drop it and rely on the 5 s cap alone, that is defensible now that the number is straight - say which you prefer.

JanZachmann:
Closing this without merging.

The only trigger we can point at is our own lab resolver, and the log does not say why it went quiet - so merging this would ship a guess to every device.

What the log does show (`test-image-job` 541): the lookup blocks for its full resolver timeout before failing - `aziot-certd` starts at 18:09:14, `Temporary failure in name resolution` at 18:09:34 - while the rest of the network was working the whole time: the test runner held ssh sessions to the DUT continuously from 18:09:15 to 18:09:34, and timesyncd had reached the lab time server at 18:09:08. So the queries went out and nothing came back from the resolver. That also rules out resolved starting with no server at all, which fails immediately instead of blocking.

Two things make the product-side wait not worth its cost:

* the failure is self-healing - identityd restarts after 5 s and the second attempt succeeds
* the one real consequence, `AducIotAgent` crashing on a failed health check, is already fixed in omnect/iot-hub-device-update#55

And the gate only ever helped when the resolver was silent for less than 30 s. Past that it starts the daemon anyway, which is exactly today's behaviour. That is a narrow win for a script plus `bash`, `coreutils` and `toml-cli` on every boot of every device.

Not solved by closing: the nightly still fails on `found syslog errors not in whitelist` when this happens. That needs a decision - either the identityd lines stay whitelisted as expected transient noise, or someone looks at why the lab resolver is silent for ~20 s after a DUT boots.

The review findings from this round were valid independently of the outcome and are worth keeping in mind for the next script of this shape: the URL parsing, the dedup array, and the `-` on `ExecStartPre` that would have hidden a broken image.

PR #684, opened 2026-08-10
fix(omnect-base-files): select the device id fallback on an empty toml result

JanZachmann:
## Summary

`omnect_get_deviceid.sh` falls back to `provisioning.device_id` on an empty toml result instead of testing for the string `null`, lets `toml get -r` print the id unquoted, and logs when it ends up with no id at all.

## Reason

`toml` (toml-cli 0.2.3, pinned by `SRCREV`) exits non-zero and prints nothing for an absent key - it never prints `null`, so the fallback was dead code. A device provisioned through `provisioning.device_id` rather than DPS attestation therefore got an empty device id, and since the script is sshd's `AuthorizedPrincipalsCommand`, no principals at all.

The quotes the script stripped came from toml-cli's JSON output, not from the value itself, so `-r` removes the need for the `tr` process.

An empty result had no signal of its own. sshd takes the output as the principals list and logs nothing when it is empty, so a certificate login failed without a reason anywhere. The stderr of `toml` is not a reliable stand-in: it is silent for an absent key, and it only reaches the journal at all because sshd asks for the subprocess stderr to be discarded while its helper acts on the stdout flag instead - a one-word upstream fix would take it away. The script says so itself now, and the `2>/dev/null` that could only hide a missing or unparsable config is gone as well.

### DIFF
--- a/recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
+++ b/recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
@@ -3,7 +3,7 @@
 # toml exits non-zero with no output for an absent key, so an empty result is
 # what selects the fallback
 device_id=$(toml get /etc/aziot/config.toml provisioning.attestation.registration_id 2>/dev/null)
-[ -n "${device_id}" ] && [ "${device_id}" != "null" ] \
+[ -n "${device_id}" ] \
     || device_id=$(toml get /etc/aziot/config.toml provisioning.device_id 2>/dev/null)
 
 echo "${device_id}" | tr -d '"'
--- a/recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
+++ b/recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-device_id=$(toml get /etc/aziot/config.toml provisioning.attestation.registration_id)
-[[ "${device_id}" == "null" ]] && device_id=$(toml get /etc/aziot/config.toml provisioning.device_id)
+# toml exits non-zero with no output for an absent key, so an empty result is
+# what selects the fallback
+device_id=$(toml get /etc/aziot/config.toml provisioning.attestation.registration_id 2>/dev/null)
+[ -n "${device_id}" ] && [ "${device_id}" != "null" ] \
+    || device_id=$(toml get /etc/aziot/config.toml provisioning.device_id 2>/dev/null)
 
-echo ${device_id} | tr -d '"'
+echo "${device_id}" | tr -d '"'
--- a/recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
+++ b/recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
-# toml exits non-zero with no output for an absent key, so an empty result is
-# what selects the fallback
+# toml prints nothing for an absent key
 device_id=$(toml get /etc/aziot/config.toml provisioning.attestation.registration_id 2>/dev/null)
 [ -n "${device_id}" ] \
     || device_id=$(toml get /etc/aziot/config.toml provisioning.device_id 2>/dev/null)
--- a/recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
+++ b/recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
@@ -5,4 +5,8 @@ device_id=$(toml get -r /etc/aziot/config.toml provisioning.attestation.registra
 [[ -n "${device_id}" ]] \
     || device_id=$(toml get -r /etc/aziot/config.toml provisioning.device_id)
 
+# sshd logs nothing of its own when the principals list comes back empty
+[[ -n "${device_id}" ]] \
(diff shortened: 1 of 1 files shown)


recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
```diff
@@ -1,6 +1,8 @@
...
 
-device_id=$(toml get /etc/aziot/config.toml provisioning.attestation.registration_id)
-[[ "${device_id}" == "null" ]] && device_id=$(toml get /etc/aziot/config.toml provisioning.device_id)
+# toml prints nothing for an absent key
+device_id=$(toml get /etc/aziot/config.toml provisioning.attestation.registration_id 2>/dev/null)
+[ -n "${device_id}" ] \
+    || device_id=$(toml get /etc/aziot/config.toml provisioning.device_id 2>/dev/null)
 
-echo ${device_id} | tr -d '"'
+echo "${device_id}" | tr -d '"'
```
HarryWaschkeit: comprehension question: where can double quotes come from so that they need to be deleted?
to my understanding, `device_id` needs to contain double quotes so that their deletion is necessary; wouldn't that be an error?
(I know that this handling isn't new but still I'd like to pose that question)
besides: double quotes could be deleted as easy as using `echo "${device_id//\"/}", no need for an extra process.

JanZachmann: The quotes come from toml-cli, not from the value: `toml get` prints JSON by default, so a TOML string comes out as `"jza-gateway-devel-rpi"`. `device_id` itself never contains quotes, so nothing is wrong with the config.

But there is no need to strip them at all - `toml get` has `-r`/`--raw`, which prints a string directly. Done in 4ff6159, which also drops the `tr` process and switches the test to `[[ ]]`.

While checking this I also found the `2>/dev/null` I had added is useless for the case it claims to cover: an absent key is a `SilentError` in toml-cli and prints nothing on stderr. The redirect only hid real errors (missing or unparsable `/etc/aziot/config.toml`), which is exactly the signal we want in the journal when sshd gets no principals - so it is gone too.

Measured on a device (toml-cli 0.2.3):

```
absent key      rc=1 stdout=[]                        stderr=[]
absent file     rc=1 stdout=[]                        stderr=[toml: No such file or directory (os error 2)]
present, plain  rc=0 stdout=["jza-gateway-devel-rpi"] stderr=[]
present, -r     rc=0 stdout=[jza-gateway-devel-rpi]   stderr=[]
```

Old vs new script, same device: DPS config gives `jza-gateway-devel-rpi` in both; a manual config with `provisioning.device_id` gives empty in the old one and `manual-device-42` in the new one.

recipes-omnect/omnect-base-files/omnect-base-files/usr/bin/omnect_get_deviceid.sh
```diff
@@ -1,6 +1,12 @@
...
 
-device_id=$(toml get /etc/aziot/config.toml provisioning.attestation.registration_id)
-[[ "${device_id}" == "null" ]] && device_id=$(toml get /etc/aziot/config.toml provisioning.device_id)
+# toml prints nothing for an absent key
+device_id=$(toml get -r /etc/aziot/config.toml provisioning.attestation.registration_id)
+[[ -n "${device_id}" ]] \
+    || device_id=$(toml get -r /etc/aziot/config.toml provisioning.device_id)
 
-echo ${device_id} | tr -d '"'
+# sshd logs nothing of its own when the principals list comes back empty
```
mlilien: what does this have to do with sshd? logger logs to syslog.

JanZachmann: This script is directly invoked by sshd as the `AuthorizedPrincipalsCommand` (configured in recipes-connectivity/openssh/openssh_%.bbappend). If it returns an empty string due to a missing device ID, sshd rejects the authentication but does not log the empty principals list. The explicit `logger` call ensures a trace is written to syslog to debug SSH connection failures in this scenario.

mlilien: the relevant part is, that it does log to syslog

JanZachmann: Imho `syslog` is not the relevant part, since it is just a detail that names the concrete logger used in `omnect-os`.

Anyway reworded in bda0619 so the comment carries the context it was missing:

```
# sshd runs this as AuthorizedPrincipalsCommand and reports only the failed
# authentication, not the empty principals list, so put the reason in syslog
```

Your first question was the right one - nothing in this file said sshd invokes it, so the old wording only made sense if you already knew that. It also overstated the case, and so did my reply above: sshd does log the failed authentication, it just does not say the principals list was empty. That cause is what the `logger` line adds.

mlilien: for the dev of this file, it is irrelevant that sshd runs it, because its not the only usecase of this script. 
the relevant part is, that it gets logged to syslog if the device_id is empty.

JanZachmann: You were right and my last reply was wrong on the facts. I checked the callers: sshd is not the only one. `wifi-commissioning-gatt@.service` uses the script for the BLE name.

```
ExecStart=/bin/bash -c '/usr/bin/wifi-commissioning-gatt-service -i %i -b $(/usr/bin/omnect_get_deviceid.sh)'
```

So naming sshd describes one of two callers instead of this file, which is what you said.

What is left to say - it gets logged when the device id is empty - is what the code says, so the comment is gone in 5f8f6b1 instead of getting a third wording.

mlilien (approved):

HarryWaschkeit (approved):

PR #687, opened 2026-08-18
feat(iotedge): update Azure IoT stack to 1.6.0 and rust toolchain to 1.97.1

mlilien:
Moves the Azure IoT stack to the 1.6.0 line, which forces a Rust toolchain bump.

Three commits, reviewable independently.

### `feat(rust): bump toolchain to 1.97.1`

oe-core wrynose (OE 6.0.2) ships rustc 1.94.1, but iotedge 1.6.0's edgelet pulls `sysinfo` 0.39.x, whose MSRV is 1.95. Bumping the toolchain was preferred over downgrading the dependency.

The rust recipes are patched in oe-core **in place** via the kas patch mechanism rather than vendored into meta-omnect: oe-core's `meta` is first in `BBPATH`, so a vendored `rust_*.bb`'s `require recipes-devtools/rust/rust-source.inc` would resolve back to oe-core's old copy.

The patch is derived from wrynose's own 1.94.1 recipe set, which is class-compatible. oe-core master's set is deliberately *not* used — it carries a work-shared source refactor (`common-source.inc`, `rust-source_1.97.1.bb`), the `@LLVM_CONFIG_PATH@`/`STAGING_BINDIR_CROSS` llvm-config rework, and `cargo_common`/`rust-common`/`cargo` class changes that would disrupt every Rust recipe in the build. It:

- renames `rust`, `cargo`, `libstd-rs` and `rust-cross-canadian` from `_1.94.1` to `_1.97.1`
- updates the source sha256 and trims the patch list in `rust-source.inc`
- moves the bootstrap snapshot from 1.93.0 to 1.96.0
- replaces `files/` with the 1.97.1 patch set

Upstream does its own bumps the same way — `git mv` plus `rust-snapshot.inc`/`rust-source.inc` metadata, with no content change to `rust_*.bb` (checked for the 1.95.0, 1.96.0 and 1.96.1 bumps). The 1.97.1 bump is the exception, so three content changes are taken from it as well:

- `rust_1.97.1.bb`: delete vendored archives from the installed library sources in all three `rust_do_install` variants. Without it `do_package` fails — 1.97 vendors the library workspace into `<src>/library/vendor`, which the recipe copies into the package, and `dwarfsrcfiles` rejects wit-bindgen 0.57.1's non-ELF archive (`oe/package.py`, `source_info()`, exempts only clang+LTO).
- `meta/classes-recipe/rust-target-config.bbclass`: the aarch64 AAPCS64 non-leaf frame pointer from the same commit, plus the arm/armv7 `eabi`/`eabihf` target ABI from the 1.96.0 bump. The arm half is not a breaker for us (rustc 1.97.1 permits hard-float with an unspecified `cfg(target_abi)`, and every arm32 machine we build is hard-float), but it belongs to the same version-bump pair and a soft-float arm32 target would be rejected outright without it.
- `meta/recipes-devtools/clang/`: `0043-LoopUnroll-Iterate-exit-fold-candidates-in-determini.patch` plus its `common.inc` `SRC_URI` line — an LLVM backport making `LoopUnroll` and `LoopPeel` iterate deterministically. Version-coupled because rust links oe-core's LLVM rather than a bundled one (`rust_1.97.1.bb` sets `download-ci-llvm = False` and points `llvm-config` at `${STAGING_BINDIR}`/`${STAGING_BINDIR_NATIVE}`), so every Rust recipe in the image draws its codegen from that tree. `llvm-project-source.bb` is the sole consumer of that patch list — it is the only recipe reaching `common.inc` through `common-clang.inc` without also requiring `common-source.inc`, which blanks `SRC_URI`. Upstream's `0043` number is kept: its `0042` comes from `789d416dd8` (`clang-tools-extra: disable tests`), which is not version-coupled and which wrynose does not carry, so keeping the number lines up on a later rebase instead of colliding.

Not adopted, deliberately: upstream's armv8a AArch32 Thumb fixes and MIPS ABI class changes (not tied to a version bump, and no machine we build needs them) and the snapshot self-contained-lld patchelf loop (only fails on build hosts without a `/lib64` compat path).

LLVM needs no version change: wrynose already ships 22.1.3 and rust 1.97 requires only >= 20. `RUSTVERSION` is retargeted so the `PREFERRED_VERSION_*` values it feeds no longer point at the removed 1.94.1.

Per the one-`p001`-per-repo convention, `kas/patches/oe.patch` is regenerated as the concatenation of `oe_layerdir.patch` and the new `oe_rust-1.97.1.patch`.

### `feat(iotedge): update Azure IoT stack to 1.6.0`

Crate manifests regenerated with `cargo bitbake`.

- **aziot-identityd** moves from the 1.6.0-rc.1 commit pinned during the wrynose migration to the 1.6.0 release tag. Its crate list came out byte-identical to rc.1, so both recipe patches still apply unchanged.
- **iotedge / aziot-edged** move to Azure/iotedge 1.6.0; version-specific includes renamed and generated recipes regenerated (crates.io set grew 226 → 306). `iot_identity_service_rev` stays at the aziot-identityd SRCREV, since iotedge 1.6.0's `edgelet/Cargo.lock` pins exactly that commit and its 26 git members are unchanged, so the remap paths in `omnect_rust_aziot-identityd_deps.bbclass` remain valid.

`RUSTFLAGS += "-D warnings"` is kept, but four patches are dropped:

- the three `fix_linter_findings-*` patches — upstream 1.6.0 absorbed all of them (`[workspace.lints.rust]` now declares the `unexpected_cfgs` check-cfg values, `External` carries `#[allow(dead_code)]`, the dead `Uid(u32)` field is gone, and `ossl110` is declared and set by openssl-build's build script)
- `fix_docker_engine_api_version.patch` — upstream moved the Docker Engine API to v1.41, compatible with the v1.40 minimum of Docker 29.5, and deleted the swagger file the patch edited. Forcing v1.44 again would regress compatibility with older daemons.

### `docs(kas): document oe.patch generation`

`p001` for openembedded-core now points at the generated `oe.patch`, so nothing references `oe_layerdir.patch` by name any more and editing it has no effect and raises no error. A new `kas/patches/README.md` plus a comment at the `p001` entry record the rule. The note is deliberately not a header inside `oe.patch`, which would break byte-equality with `cat kas/patches/oe_*.patch`.

### Verification

- `oe.patch` byte-equals `cat kas/patches/oe_*.patch`, and applies to a pristine `5d1aa5c806` worktree with `git apply --check --whitespace=nowarn`, exactly how kas applies it.
- `0043-LoopUnroll-…` is byte-identical to upstream's copy, and was test-applied against the `llvm-project-22.1.3.src` tarball `${BASEURI}` resolves to: both hunks land at the line numbers in the patch header (`LoopPeel.cpp` 11 and 1122, `LoopUnroll.cpp` 17 and 516) with `git apply --check` clean, so no fuzz and no offset. None of the pinned `clang/` patch files touch either source file, so its position after `0041` rather than upstream's `0042` is immaterial. After applying, `common.inc` differs from upstream master only by the SPDX license string, the 22.1.8 `sha256sum` and `0042`.
- Every file the patch produces was diffed against oe-core master's *same-version* file, and each residual hunk accounted for: `rust-snapshot.inc` is byte-identical, `rust-source.inc` differs only by the work-shared block, `rust_1.97.1.bb` and `rust-target-config.bbclass` only by the refactors and the deliberately-skipped hunks listed above.


HarryWaschkeit (approved):

kas/patches/oe_rust-1.97.1.patch
```diff
@@ -0,0 +1,873 @@
...
+-            file://0001-Adjust-loongarch-assembly-test.patch;patchdir=${RUSTSRC} \
+-            file://0001-Fix-multiple-option-or-permutations-test-for-big-end.patch;patchdir=${RUSTSRC} \
++            file://0002-Fix-rust-build-failure-with-unstable-options.patch;patchdir=${RUSTSRC} \
+ "
+-SRC_URI[rust.sha256sum] = "174fce10ce012317ca995810296d8af199318838180b03d68a853e0f02d4b571"
++SRC_URI[rust.sha256sum] = "0ed06fdaffd4722a7702e0b4eebfafc897ab8f513e8e1b247cdd7e5c6df6ded2"
+ 
+ RUSTSRC = "${UNPACKDIR}/rustc-${RUST_VERSION}-src"
+ 
+diff --git a/meta/recipes-devtools/rust/rust_1.94.1.bb b/meta/recipes-devtools/rust/rust_1.97.1.bb
```
JanZachmann: This rename keeps the recipe byte-identical, so it misses the vendored-archive cleanup that oe-core added in the same 1.97.1 bump (`65271a5d`). Upstream `rust_1.97.1.bb` has this line after each of the three `cp -r ${S}/library` sites:

```
find ${D}${libdir}/rustlib/src/rust/ -path "*/vendor/*.a" -type f -delete
```

Why it matters: at 1.97.1 the bootstrap vendors the library workspace into `<src>/library/vendor`, while 1.94.1 vendored everything into one top-level `vendor/`. `library/Cargo.lock` at 1.97.1 pins `wit-bindgen 0.57.1`, whose crate ships `src/rt/libwit_bindgen_cabi.a`, a wasm archive. All three `rust_do_install` variants do `cp -r ${S}/library ${D}${libdir}/rustlib/src/rust`, so that archive now lands in the packaged sources, and `oe.package.source_info()` runs `dwarfsrcfiles` on static libs and calls `bb.fatal` on a non-ELF one (only clang + LTO is exempt), which fails `do_package`. At 1.94.1 the archive sat in `${S}/vendor`, which is not copied, so this only breaks after the bump.

Please add the upstream `do_install` hunk at all three sites.

kas/patches/oe_rust-1.97.1.patch
```diff
@@ -0,0 +1,873 @@
+diff --git a/meta/recipes-devtools/rust/cargo_1.94.1.bb b/meta/recipes-devtools/rust/cargo_1.97.1.bb
```
JanZachmann: The patch scope is `meta/recipes-devtools/rust/` only, but the same upstream 1.97.1 commit (`65271a5d`) also changed `meta/classes-recipe/rust-target-config.bbclass`, and the 1.96 bump changed it too.

Concrete gap for us: upstream sets `tspec['frame-pointer'] = "non-leaf"` for aarch64 (AAPCS64), the pinned OE 6.0.2 copy does not. The generated aarch64 target JSON then differs from rustc's built-in `aarch64-unknown-linux-gnu`, which affects codegen and backtrace quality on the aarch64 machines.

The 1.96 half (`abi = "eabi"/"eabihf"` for arm/armv7) is not a breaker: rustc 1.97.1 `check_consistency` allows `(Some(FloatAbi::Hard), CfgAbi::Unspecified)`, and every omnect arm32 machine is hard-float (`cortexa7thf-neon-vfpv4`). Only a soft-float ARM target would be rejected, and we build none.

kas/distro/oe.yaml
```diff
@@ -20,6 +20,6 @@ repos:
     patches:
       p001:
         repo: "meta-omnect"
-        path: "kas/patches/oe_layerdir.patch"
+        path: "kas/patches/oe.patch"
```
JanZachmann: After this switch nothing references `kas/patches/oe_layerdir.patch` any more (`git grep oe_layerdir` in `kas/` and `ci/` finds nothing), but the file stays in the tree, and `oe.patch` is a byte-identical copy of it plus the rust part (verified: `cat oe_layerdir.patch oe_rust-1.97.1.patch` equals `oe.patch`).

Every other layer keeps a directly referenced `*_layerdir.patch`, so the next person edits `oe_layerdir.patch` the usual way and gets no effect and no error. Either drop the file, or note at the top of `oe.patch` how it is generated.

JanZachmann (commented):
One likely build breaker in the rust 1.97.1 rename, plus two smaller things, all inline. One more finding cannot be inline because the file is not in the diff: `doc/examples/iotedge-apparmor/deployment.json:18` still pins `mcr.microsoft.com/azureiotedge-agent:1.5`, while 1.6.0 ships `image = "mcr.microsoft.com/azureiotedge-agent:1.6"` in `edgelet/contrib/config/linux/default.toml` (installed as `config.toml.default`, and the same value is the commented default in `config.toml.edge.template`). Worth bumping the example in this PR.

What I checked and found correct: rust source and all snapshot checksums against `static.rust-lang.org`; the patch applies cleanly on wrynose and matches oe-core 1.97.1 file for file; `RUSTVERSION ?= "1.97.1%"` wins over `tcmode-default.inc`; both `SRCREV`s equal the upstream 1.6.0 tags and match `edgelet/Cargo.lock`; the generated manifests match the locks exactly (306 crates and 306 checksums each, plus the 26 git members the `.inc` files override); all retained patches still apply and every dropped-patch claim holds at the new revisions. Dropping `fix_docker_engine_api_version.patch` is safe: iotedge 1.6.0 asks for `/v1.41` and docker-moby v29.3.0 has `defaultMinAPIVersion = "1.40"`.

For the release note: the installed default edge agent image moves to `mcr.microsoft.com/azureiotedge-agent:1.6`.

kas/distro/oe.yaml
mlilien: on purpose. oe.patch is a concatenation off all oe_*.patch files. please instruct your agent with that fact.

kas/patches/oe_rust-1.97.1.patch
mlilien: Added at all three `rust_do_install` sites, byte-identical to `65271a5d` — including upstream's asymmetry at the `class-target` site, where the existing `*.sh` line has no trailing slash on the path and the new one does.

I also ran the check that should have been part of building this patch in the first place: diff every resulting file against upstream's *same-version* file and account for every residual hunk. For `rust_1.97.1.bb` what is left is the work-shared source refactor, the `STAGING_BINDIR_CROSS` llvm-config change, the `DEPENDS` cleanup from `32c44d64` and the snapshot lld loop from `9db02003` — none of them version-specific. `rust-snapshot.inc` is byte-identical to upstream's, and `rust-source.inc` differs only by the work-shared block. So this hunk was the only real gap.

Squashed into `feat(rust): bump toolchain to 1.97.1`. Still unbuilt — `do_package` for `rust` would have been the first thing to hit this.

kas/patches/oe_rust-1.97.1.patch
mlilien: Correct, the patch scope was too narrow. Extended it to `meta/classes-recipe/rust-target-config.bbclass` with both version-coupled hunks: the aarch64 non-leaf frame pointer from `65271a5d` and the arm/armv7 `eabi`/`eabihf` ABI from `b23ee2d0`.

Your reading of the arm half holds — rustc 1.97.1 `compiler/rustc_target/src/spec/mod.rs`, `Arch::Arm` in `check_consistency`, permits `(Some(FloatAbi::Hard), CfgAbi::EabiHf | Uwp | Unspecified | Other(_))`, and every arm32 machine we build is hard-float (`raspberrypi3` → `cortexa7thf-neon-vfpv4`, dehndetect i.MX6 likewise). One thing worth adding: that `match self.arch` block sits *after* the `if kind == TargetKind::Builtin` gate closes, so it does apply to JSON targets. Only `(Some(FloatAbi::Soft), CfgAbi::Eabi)` is accepted on the soft-float side, so a soft-float arm32 machine would be rejected outright rather than merely inconsistent. That is the second reason to take the hunk now instead of when someone adds such a board.

Skipped the two class changes that are not tied to a version bump, since they retire on a different schedule and no machine we build needs them: `1808270d` (armv8a AArch32 Thumb fixes — we have no 32-bit ARMv8 machine; rpi3 is armv7a/cortexa7) and `8042c165` (MIPS ABI).

kas/distro/oe.yaml
mlilien: Confirmed — `p001` pointed at `oe_layerdir.patch` directly until this PR repointed it at `oe.patch`.

Kept the file rather than dropping it: the convention is one `p001` per repo with `oe.patch` as the tracked concatenation of the `oe_*` sources, which is what keeps those sources individually reviewable and retirable (`oe_layerdir.patch` is permanent, `oe_rust-1.97.1.patch` goes away when oe-core ships rust >= 1.97.1). Documented the rule where both readers look instead: a new `kas/patches/README.md`, and a comment at the `p001` entry in `kas/distro/oe.yaml`.

Deliberately *not* a header inside `oe.patch` — that would break byte-equality with `cat kas/patches/oe_*.patch`, which is how the generated file is verified.

Separate commit: `docs(kas): document oe.patch generation`.

kas/patches/oe_rust-1.97.1.patch
JanZachmann: Verified independently. The generated `oe.patch` — which on this head is byte-identical to `cat kas/patches/oe_*.patch`, so it is what kas really applies — applies clean at the pinned commit `5d1aa5c8`, the three hunks match `65271a5d` byte for byte (including the asymmetric trailing slash at the `class-target` site), and after applying, every file under `meta/recipes-devtools/rust/` equals upstream master's except the later refactors that are not version-coupled: the work-shared source tree (`rust-source.inc`, the `common-source.inc` requires and `CARGO_HOME` in cargo and libstd-rs), the SPDX license string normalization, and the `libgit2`/`sqlite3` `DEPENDS` change in cargo. `rust-snapshot.inc`, `rust-cross-canadian*` and all six files in `files/` are byte-identical.

The four dropped patches are right as well: `65271a5d` itself drops `0001-riscv32-…`, `0003-explicit-tail-calls-disable-two-tests`, `0004-explicit-tail-calls-ignore-tests-…` and `0005-remap-OUT_DIR-paths-…` as merged upstream, so their absence is not a gap.

Agreed on the build: `do_package` for `rust` is exactly where this hunk lands, so one aarch64 and one arm32 machine should be built before merge.

kas/patches/oe_rust-1.97.1.patch
JanZachmann: The bbclass part is right. After applying, `rust-target-config.bbclass` differs from upstream master only by `1808270d` (armv8a AArch32 Thumb: `target_arch`, `+v8`, `+thumb2`, the `rust_sys_to_llvm_target(sys, d)` signature) and `8042c165` (MIPS `abi`/`llvm-abiname`), which are the two skips you named. The aarch64 and arm hunks are byte-identical to upstream, and `rust_sys_to_llvm_target` stays 1-arg together with its only call site, so skipping `1808270d` is self-consistent.

One more piece of `65271a5d` is still out of scope, and this one is version-coupled: the same commit adds `0043-LoopUnroll-Iterate-exit-fold-candidates-in-determini.patch` to `meta/recipes-devtools/clang/common.inc`, described in the commit message as a backport "to fix the repro issue". It reaches our build: `rust_1.97.1.bb` has `DEPENDS += "llvm"`, and `common.inc` is where the `SRC_URI` patch list for the shared llvm-project source lives (`llvm_git.bb` deltasks its own `do_patch`, `llvm-project-source.bb` keeps it and both reach `common.inc` through `common-clang.inc`). Pinned 6.0.2 is at LLVM 22.1.3 with the list ending at `0041`; upstream master is at 22.1.8 and still carries `0043`, so the 22.1.x series has not absorbed it and it should still apply — but it needs a test-apply, not a blind copy.

Whether we want it is a judgement call: it fixes non-deterministic codegen, not wrong codegen. Worth deciding rather than missing.

The rest of `65271a5d` needs nothing. `tcmode-default.inc` is covered by the `RUSTVERSION ?= "1.97.1%"` override, and that value is exactly what upstream sets. `maintainers.inc` and `rust-source_1.97.1.bb` have no counterpart in 6.0.2. `oeqa/selftest/cases/rust.py` only matters when running oe-selftest.

kas/distro/oe.yaml
JanZachmann: Understood, and the invariant holds on this head: `cmp <(cat kas/patches/oe_*.patch) kas/patches/oe.patch` is identical. Withdrawing the finding.

Two small things on the README:

The "Current sources" table restates `ls kas/patches/oe_*.patch`, so it needs an edit every time a patch is added or dropped, and only the "retires when" column says something the file list does not. Moving that column into each patch's own header, or dropping the file column, would age better.

The rule is written down but nothing checks it. If someone edits an `oe_*.patch` and forgets to regenerate, kas builds the stale concatenation and nothing reports it. The repo already runs one GitHub workflow (`conventional-commit-checker.yml`), so a second job doing `cmp <(LC_ALL=C cat kas/patches/oe_*.patch) kas/patches/oe.patch` would turn "regenerate by hand" into something enforced. `LC_ALL=C` because the README makes glob order the apply order, and glob order is locale-dependent.

JanZachmann (commented):
Round 2. Both round-1 blockers are fixed and I re-checked them against upstream rather than against the description: the generated `oe.patch` applies clean at the pinned oe-core commit `5d1aa5c8`, and after applying, `meta/recipes-devtools/rust/` matches upstream master except the later refactors that are not tied to the version bump, while `rust-target-config.bbclass` differs from master by exactly the two commits you named as skipped. Details in the threads; the `vendor/*.a` one is resolved.

Three things are still open, none of them a reason to hold the code as written:

The clang/LLVM `0043-LoopUnroll-…` patch from the same upstream commit is still outside the patch scope, and it does reach this build through `llvm-project-source`. It fixes non-deterministic codegen, not wrong codegen, so it is a decision rather than a defect - see the scope thread.

`doc/examples/iotedge-apparmor/deployment.json:18` still pins `azureiotedge-agent:1.5` while 1.6.0 ships `:1.6` as its default. Outside the diff, so there is no thread for it.

The branch is unbuilt. Both new rust hunks land in `do_package` and in the generated target JSON, so a build of one aarch64 and one arm32 machine is the verification that counts here. That is why this is a comment and not an approval - the code matches upstream, but nothing has executed it yet.

kas/patches/oe_rust-1.97.1.patch
mlilien: Taken. Added `0043-LoopUnroll-Iterate-exit-fold-candidates-in-determini.patch` and its `common.inc` line byte-identical to `65271a5d`, squashed into `feat(rust): bump toolchain to 1.97.1`.

Test-applied rather than copied. Against the `llvm-project-22.1.3.src` tarball the recipe actually fetches, both hunks land at exactly the line numbers in the patch header — `LoopPeel.cpp` 11 and 1122, `LoopUnroll.cpp` 17 and 516 — and `git apply --check` is clean, so no fuzz and no offset. None of our `0001`–`0041` touch either file, so sitting after `0041` instead of after upstream's `0042` changes nothing. After applying, `common.inc` differs from master only by the SPDX license string, the 22.1.8 `sha256sum` and `0042` itself.

Two refinements to the mechanism. `common-source.inc` does more than deltask `do_patch` — it also sets `SRC_URI = ""`, and all ten other recipes in that directory require it. `llvm-project-source.bb` is the only recipe that reaches `common.inc` through `common-clang.inc` *without* it, and that is what makes it the sole consumer of the patch list.

The reach is also wider than `DEPENDS += "llvm"` alone shows: `rust_1.97.1.bb` sets `download-ci-llvm = False` and points `llvm-config` at `${STAGING_BINDIR}` / `${STAGING_BINDIR_NATIVE}`, so rustc links our LLVM 22.1.3 rather than a bundled one. Every Rust recipe in the image draws its codegen from the tree this patch fixes — omnect-device-service, the aziot services, bindgen.

That decides the judgement call for me. `MapVector` is `DenseMap` plus insertion order, so it cannot make codegen wrong; the cost is one patch; and the alternative is knowingly shipping a non-deterministic optimizer under the entire Rust payload of an update product, where a rebuild that yields different bytes is a spurious update delta.

On naming: kept upstream's `0043` rather than renumbering to `0042`. Upstream's `0042` comes from `789d416dd8` (`clang-tools-extra: disable tests`), which is not version-coupled, so keeping the number means a later rebase onto an oe-core carrying it lines up instead of colliding.

JanZachmann: Verified, including the parts I could not check last round.

`0043-LoopUnroll-…` is byte-identical to upstream's copy. I test-applied it against the real 22.1.3 sources — `LoopPeel.cpp` and `LoopUnroll.cpp` fetched at `llvmorg-22.1.3`, which is what `${BASEURI}` resolves to — and `git apply --check` is clean with no offset, so the header line numbers hold against the version we actually fetch. None of the 39 patch files in the pinned `clang/` directory mentions either file. After applying, `common.inc` differs from upstream master only by the SPDX license string, the 22.1.8 `sha256sum` and `0042`, exactly as you describe. `oe.patch` still applies clean at `5d1aa5c8` and is still byte-identical to `cat kas/patches/oe_*.patch`, and the rust half is unchanged from what I checked last round.

Your correction on the mechanism is right: 10 recipes require `common-source.inc`, 11 require `common-clang.inc`, and `llvm-project-source.bb` is the only one in the second set that is not in the first — so it is the sole consumer of the patch list. Keeping upstream's `0043` number is the right call for the same reason you give.

One loose end, in the message rather than the code: `feat(rust): bump toolchain to 1.97.1` still says "LLVM needs no change; wrynose already ships 22.1.3 and rust 1.97 requires >= 20" and "Two content changes are taken from upstream on top of the rename". The commit now carries a third one, and it is an LLVM change. Worth fixing while the branch is still being rewritten, since that text is what ends up in the history.

kas/distro/oe.yaml
mlilien: Both taken, in `d8a4ce1`.

The table is gone. It restated `ls kas/patches/oe_*.patch` and needed an edit on every add or drop, and this README is meant to read as an example of the convention rather than a registry of it. The one thing the table carried that the file listing did not - the retirement condition - survives as prose.

The check is `.github/workflows/kas-patch-checker.yml`, running the same `cmp` on every pull request against main, kirkstone, scarthgap and wrynose. It is green on this head, so it demonstrates itself.

One correction to the snippet, because it matters for exactly the reason you raised it. `cmp <(LC_ALL=C cat kas/patches/oe_*.patch)` puts `LC_ALL` in `cat`'s environment, but the glob is expanded and collated by the *shell* before `cat` is ever invoked, so the ordering you are guarding against has already been decided. It has to be set on the shell instead:

    export LC_ALL=C
    cmp <(cat kas/patches/oe_*.patch) kas/patches/oe.patch

The concern is real - with sources named `oe_A`/`oe_a`/`oe_B`, C collation yields `oe_A oe_B oe_a` and en_US.UTF-8 yields `oe_a oe_A oe_B`. The workflow therefore sets `LC_ALL: C` as step-level `env:` so the whole shell inherits it, and the README now shows the `export` form.

I also confirmed the check can fail and not merely pass: a source edited without regenerating, a new source added without regenerating, and a hand-tampered `oe.patch` are each caught.

mlilien:
Round 2 items closed.

**Builds.** One aarch64 machine (`raspberrypi4-64`) and one arm32 machine (`dehndetect`, armv7 hard-float i.MX6) both build green, image included. Both pull in `aziot-edged 1.6.0`, `aziot-identityd`, `iotedge 1.6.0` and `omnect-device-service`, so the iotedge half of the PR is covered and not just the toolchain. The split is deliberate: the aarch64 frame-pointer hunk and the arm/armv7 `eabihf` hunk each only show up in the target JSON of their own architecture, so one machine could not have exercised both.

**One correction on your `do_package` point.** "Both new rust hunks land in `do_package`" holds for the generated target JSON but not for the `vendor/*.a` deletion. Target `rust` is not built by our images at all - only `rust-native`, `cargo-native` and target `libstd-rs`. Nothing RDEPENDs on target rust, and meta-omnect ships no toolchain kas feature, so `do_package` for `rust` never runs and that hunk is not exercised by an image build.

It is still the right hunk to carry: `library/vendor/wit-bindgen-0.57.1/src/rt/libwit_bindgen_cabi.a` is present in the 1.97.1 source, inside the `library/` tree that `rust_do_install` copies, so the hazard is real for anyone who builds target rust or an SDK. Treat that one as reasoned rather than executed - say so if you want it actually built.

**`deployment.json`** is fixed in `f350428`, and it was three pins rather than the one you cited: `azureiotedge-agent`, `azureiotedge-hub` and the SimulatedTemperatureSensor sample all sat at `:1.5`. agent and hub are the runtime pair and Azure does not support a 1.6 agent against a 1.5 hub, so changing only line 18 would have left the manifest worse than it was. Bare minor tags kept, matching the file; all three `1.6` tags exist in MCR.

**Scope item**: `0043` landed before this review, in `75efeaf`. The stale "not adopted, deliberately" line and the two-vs-three count in the PR body were corrected in the same push.

The README and CI-enforcement points from the `kas/distro/oe.yaml` thread are answered there, in `d8a4ce1`.

One caveat on reproducing the arm32 result: dehndetect pulls `meta-omnect-dehn` @ `73c1eea` and `mlilien/meta-dehn` @ `965ca1a`, both still on personal forks pending omnect/meta-dehn#11 and omnect/meta-omnect-dehn#40.

kas/patches/oe_rust-1.97.1.patch
JanZachmann: Fixed — "Three content changes", the LLVM sentence now reads "no version bump", and the third change has its own paragraph.

One correction to what I wrote, because it changes where accuracy matters: I said that commit text is what ends up in the history. It is not. This repo is squash-only (`allow_merge_commit` and `allow_rebase_merge` are both false) with `squash_merge_commit_title = PR_TITLE` and `squash_merge_commit_message = PR_BODY` — which is also why the commit linter runs with `ignoreCommits: true`. So `main` records the PR title and body, and the per-commit messages live on the branch and in review only. The body already had this right, so nothing further here.

kas/distro/oe.yaml
JanZachmann: You are right about `LC_ALL=C cat` — the glob is expanded and collated by the shell before `cat` ever gets that environment, so my snippet guarded nothing. Step-level `env:` is the correct place, and the README's `export` form matches what the workflow does.

The check itself does what it claims: `shell: bash` for the process substitution, `LC_ALL: C` in the step environment so the collation applies to the glob, and `cmp` of the concatenation against the tracked file. `kas patches` is green on this head, and I get the same result locally.

One limit worth knowing rather than fixing: it covers the pull-request path only, so a push straight to a target branch is not checked. Fine as long as those branches stay PR-only.

JanZachmann (approved):
Approving. All three threads are resolved and nothing is open from my side.

What I checked against upstream rather than against the description, across the rounds: the generated `oe.patch` applies clean at the pinned oe-core commit `5d1aa5c8` and stays byte-identical to `cat kas/patches/oe_*.patch`; after applying, `meta/recipes-devtools/rust/` matches upstream master except the refactors that are not tied to the version bump, `rust-target-config.bbclass` differs only by the two class commits you deliberately skipped, and `common.inc` only by the SPDX string, the 22.1.8 `sha256sum` and `0042`. `0043-LoopUnroll-…` is byte-identical to upstream and applies with no offset against the real `llvmorg-22.1.3` sources. On the iotedge side, both `SRCREV`s equal the upstream 1.6.0 tags, the generated manifests match the locks exactly, every retained patch still applies and every dropped one is genuinely absorbed upstream, and all three example images have a `1.6` tag on MCR for amd64, arm32v7 and arm64v8.

The build result and the CI green are yours, not mine - I did not run the image build.

.github/workflows/kas-patch-checker.yml
```diff
@@ -0,0 +1,29 @@
...
+      - reopened
+jobs:
+  oe-patch-up-to-date:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: oe.patch matches its oe_* sources
+        shell: bash
+        # LC_ALL belongs in the shell environment, not on cat: the glob is expanded and
+        # collated before cat runs, and glob order is the order the hunks get applied.
```
HarryWaschkeit: that seems a leftover from the checkers evolution; is it really necessary?

mlilien: you mean the comment itself?

mlilien: removed comment

HarryWaschkeit (approved):

PR #688, opened 2026-08-19
chore: drop unused image files and the perl build dependency, restrict /dev/crypto

JanZachmann:
## Summary

- **openssl** bbappend: delete `c_rehash` and `${libdir}/ssl-3/misc`
- **nss** bbappend: restrict `FILES:${PN}` to the libraries, the config and `shlibsign`, and move the rest of `${bindir}` into a new `nss-tools` package that no image installs
- **omnect-os-distro.conf**: `RDEPENDS:nspr-dev:remove = "perl"`
- **cryptodev-udev**: new file-only recipe shipping `10-cryptodev.rules`, which gives `/dev/crypto` mode `0600`; installed on tauril2 through the machine include. The module itself stays in the image.

## Reason

`omnect-os-distro.conf` already drops `openssl-misc` and `nss-smime` from `PACKAGES` to keep perl out of the image, but `PACKAGES:remove` only removes the package - the files move into the respective main package and still ship, as perl scripts in an image without a perl interpreter.

nss is the bigger case: the recipe has `FILES:${PN} = "... ${bindir} ..."`, so **49 tools** ship (`certutil`, `pk12util`, `selfserv`, `tstclnt`, `ssltap`, `signtool`, the test binaries, ...) although nothing in the image references any of them - networkmanager pulls nss in for the libraries only. Restricting the package and collecting the tools in `nss-tools` keeps them buildable for a debug image and takes them out of the release image, including `smime`, so no single-file removals are needed there. `shlibsign` stays in the main package because the nss postinst calls it, and that postinst can end up running on first boot. openssl keeps the file removal: upstream splits those two paths into `openssl-misc` and the main package claims them only because the distro drops that package.

perl itself shows up in the image SBOM although no perl file is in the image: the SBOM is written per built recipe, and the only recipe pulling target perl into the build is nspr (`RDEPENDS:${PN}-dev += "perl"`, nspr comes in via `networkmanager -> nss -> nspr`). Dev packages are never installed, so dropping that dependency changes nothing in the image while taking perl out of the build and out of the SBOM - four high/critical findings less to moderate per release.

On tauril2 the imx BSP installs the cryptodev module, it is loaded at boot, and the driver creates `/dev/crypto` world read/write, while upstream has no fix for CVE-2026-28529 (use-after-free in `get_userbuf()`, local privilege escalation). The module stays available for applications; the udev rule takes away the world read/write default, and its low number lets a customer rule override the mode.

No other BSP ships cryptodev, and no script, unit or binary in the 6.0.2 images references the removed files or `/dev/crypto`.

### DIFF
--- a/conf/machine/include/phytec-imx8mm.inc
+++ b/conf/machine/include/phytec-imx8mm.inc
@@ -11,12 +11,6 @@ IMAGE_CLASSES:remove = " wic-helper wic-imx8-helper"
 # don't install imx-m4-demos
 MACHINE_EXTRA_RRECOMMENDS:remove = "imx-m4-demos"
 
-# don't install the cryptodev kernel module: nothing in the image opens
-# /dev/crypto (there is no openssl cryptodev engine), the driver registers the
-# device node world read/write, and upstream cryptodev-linux has no fix for the
-# page reference bug in get_userbuf() (CVE-2026-28529).
-BAD_RECOMMENDATIONS += "kernel-module-cryptodev"
-
 # install kernel-devicetree to 'root' partition instead of 'boot' and only install boot scripts to boot
 IMAGE_BOOT_FILES:remove:pn-omnect-os-image = " \
     ${KERNEL_DEVICETREE_FN} \
@@ -28,6 +22,9 @@ IMAGE_BOOT_FILES:remove:pn-omnect-os-image = " \
 "
 
 IMAGE_INSTALL:append:pn-omnect-os-image = " kernel-devicetree"
+
+# the imx BSP installs the cryptodev module; restrict its device node to root
+IMAGE_INSTALL:append:pn-omnect-os-image = " cryptodev-udev"
 
 # initramfs type
 OMNECT_INITRAMFS_FSTYPE = "cpio.gz.u-boot"
--- a/conf/machine/include/phytec-imx8mm.inc
+++ b/conf/machine/include/phytec-imx8mm.inc
@@ -11,6 +11,12 @@ IMAGE_CLASSES:remove = " wic-helper wic-imx8-helper"
 # don't install imx-m4-demos
 MACHINE_EXTRA_RRECOMMENDS:remove = "imx-m4-demos"
 
+# don't install the cryptodev kernel module: nothing in the image opens
+# /dev/crypto (there is no openssl cryptodev engine), the driver registers the
+# device node world read/write, and upstream cryptodev-linux has no fix for the
+# page reference bug in get_userbuf() (CVE-2026-28529).
+BAD_RECOMMENDATIONS += "kernel-module-cryptodev"
+
 # install kernel-devicetree to 'root' partition instead of 'boot' and only install boot scripts to boot
 IMAGE_BOOT_FILES:remove:pn-omnect-os-image = " \
     ${KERNEL_DEVICETREE_FN} \
--- a/recipes-support/nspr/nspr_%.bbappend
+++ b/recipes-support/nspr/nspr_%.bbappend
@@ -1,0 +1,5 @@
+# nspr-dev is the only recipe in the image build that requires target perl, and we
+# never install dev packages. Dropping the dependency keeps the perl recipe out of
+# the build and out of the SBOM, where it otherwise shows up as a shipped
+# component although no perl file is in the image.
+RDEPENDS:${PN}-dev:remove = "perl"
--- a/recipes-support/nspr/nspr_%.bbappend
+++ b/recipes-support/nspr/nspr_%.bbappend
@@ -1,5 +1,0 @@
-# nspr-dev is the only recipe in the image build that requires target perl, and we
-# never install dev packages. Dropping the dependency keeps the perl recipe out of
-# the build and out of the SBOM, where it otherwise shows up as a shipped
-# component although no perl file is in the image.
-RDEPENDS:${PN}-dev:remove = "perl"
--- a/recipes-support/nss/nss_%.bbappend
+++ b/recipes-support/nss/nss_%.bbappend
@@ -1,0 +1,11 @@
+# nss packages the whole ${bindir}, so its test and demo tools ship in the image.
+# Drop the ones that only exist for testing, plus smime, which is perl and has no
+# interpreter in the image (nss-smime is dropped from PACKAGES in
+# omnect-os-distro.conf, which leaves the file in the main package).
+do_install:append:class-target() {
+    rm -f ${D}${bindir}/bltest \
+          ${D}${bindir}/dbtool \
+          ${D}${bindir}/ecperf \
+          ${D}${bindir}/fipstest \
+          ${D}${bindir}/smime
+}
--- a/recipes-support/nss/nss_%.bbappend
+++ b/recipes-support/nss/nss_%.bbappend
@@ -1,8 +1,12 @@
-# nss packages the whole bindir; drop the test tools and the perl script smime.
-do_install:append:class-target() {
-    rm -f ${D}${bindir}/bltest \
-          ${D}${bindir}/dbtool \
-          ${D}${bindir}/ecperf \
-          ${D}${bindir}/fipstest \
-          ${D}${bindir}/smime
-}
+# nss packages the whole bindir, but the image uses none of the tools. Keep
+# shlibsign, which the postinst needs if it runs on first boot, and move the rest
+# into a package that is not installed.
+PACKAGES:append:class-target = " ${PN}-tools"
+
+FILES:${PN}:class-target = " \
+    ${sysconfdir} \
+    ${bindir}/shlibsign \
+    ${libdir}/lib*.chk \
+    ${libdir}/lib*.so \
+"
+FILES:${PN}-tools = "${bindir}"
--- a/recipes-support/nss/nss_%.bbappend
+++ b/recipes-support/nss/nss_%.bbappend
@@ -1,7 +1,4 @@
-# nss packages the whole ${bindir}, so its test and demo tools ship in the image.
-# Drop the ones that only exist for testing, plus smime, which is perl and has no
-# interpreter in the image (nss-smime is dropped from PACKAGES in
-# omnect-os-distro.conf, which leaves the file in the main package).
+# nss packages the whole bindir; drop the test tools and the perl script smime.
 do_install:append:class-target() {
     rm -f ${D}${bindir}/bltest \
           ${D}${bindir}/dbtool \
--- a/conf/distro/include/omnect-os-distro.conf
+++ b/conf/distro/include/omnect-os-distro.conf
@@ -163,6 +163,8 @@ PACKAGECONFIG:pn-opkg-utils = "update-alternatives"
 # get rid of perl as runtime dependency in release image
 PACKAGES:remove:pn-nss = "nss-smime"
 PACKAGES:remove:pn-openssl = "openssl-misc"
+# nspr-dev is the only recipe pulling target perl into the build
+RDEPENDS:nspr-dev:remove = "perl"
 PACKAGECONFIG:remove:pn-apparmor = "perl"
 DEPENDS:remove:pn-apparmor = "perl"
 # apparmor inherits cpan, which exports PERLCONFIGTARGET=yes and makes the native
--- a/recipes-connectivity/openssl/openssl_%.bbappend
+++ b/recipes-connectivity/openssl/openssl_%.bbappend
@@ -1,7 +1,5 @@
-# openssl-misc is dropped from PACKAGES (see omnect-os-distro.conf) to keep perl
-# out of the image. That only removes the package, not its files - they end up in
-# the main openssl package instead. c_rehash and the misc scripts are perl, so
-# they cannot run in an image without an interpreter.
+# openssl-misc is dropped from PACKAGES, which leaves its perl scripts in the
+# main package - remove the files as well.
 do_install:append:class-target() {
     rm -f ${D}${bindir}/c_rehash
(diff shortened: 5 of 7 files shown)


JanZachmann:
@mlilien two things I could not settle myself, both need your call rather than mine.

**1. I could not build, so three points are on CI.** Please check them before merging:

- the files are actually gone from the images and no `installed-vs-shipped` QA warning shows up for openssl or nss (deleting in `do_install` should avoid it, but that is the one thing a bbappend like this typically gets wrong);
- `cryptodev.ko` and `modules-load.d/cryptodev.conf` are gone from the tauril2 image. `BAD_RECOMMENDATIONS` only catches a *recommendation*. If the i.MX BSP pulls the module in as a hard dependency somewhere, it will still be in the image and the change needs a different mechanism - I could not find the place that adds it from the release sources, only that `meta-freescale` sets `MACHINE_EXTRA_RRECOMMENDS = "kernel-modules"`;
- the tauril2 smoke tests stay green, in case a hardware-crypto path did use `/dev/crypto` after all. My check was static: no file under `/usr` references the device and `/usr/lib/engines-3/` has only `aziot_keys.so`, so there is no OpenSSL cryptodev engine - but "no reference in the rootfs" is not the same as "nothing needs it".

**2. nss ships about 35 more tools that are equally unused.** Besides the four test binaries this PR removes, the image carries `certutil`, `pk12util`, `modutil`, `crlutil`, `selfserv`, `strsclnt`, `tstclnt`, `ssltap`, `signtool`, `signver`, `vfychain`, `vfyserv`, `atob`, `btoa`, `derdump`, `pp`, `addbuiltin`, `makepqg`, `ocspclnt`, `oidcalc`, `p7content`, `p7env`, `p7sign`, `p7verify`, `pwdecrypt`, `rsaperf`, `sdrtest`, `symkeyutil`, `dbtest`, `multinit`, `nonspr10`, `baddbdir`, `encodeinttest`, `listsuites`, `pkix-errcodes`, `nss-policy-check` and `shlibsign` - the recipe has `FILES:${PN} = "... ${bindir} ..."`, so everything it builds ships. NSS itself is only in the image as a transitive dependency of networkmanager, which uses the libraries, not the tools.

I left them in on purpose: dropping the whole tool set is a bigger decision than this PR, and `shlibsign` is called from the nss postinst, so it cannot simply go. Two options - tell me which you prefer and I do it:

- keep the tools (status quo of this PR), or
- restrict `FILES:${PN}` to the libraries plus `shlibsign` (and `certutil` if you want a way to inspect the cert store on a device), which would remove roughly 35 binaries from every image.

conf/machine/include/phytec-imx8mm.inc
```diff
@@ -11,6 +11,12 @@ IMAGE_CLASSES:remove = " wic-helper wic-imx8-helper"
 # don't install imx-m4-demos
 MACHINE_EXTRA_RRECOMMENDS:remove = "imx-m4-demos"
 
+# don't install the cryptodev kernel module: nothing in the image opens
+# /dev/crypto (there is no openssl cryptodev engine), the driver registers the
+# device node world read/write, and upstream cryptodev-linux has no fix for the
+# page reference bug in get_userbuf() (CVE-2026-28529).
+BAD_RECOMMENDATIONS += "kernel-module-cryptodev"
```
mlilien: a user could potentially use it in an application, so i would not do that, or you have to monitor openssl if there is a fix and reenable it in the futurre.

JanZachmann: Fair point, and I would not overrule it - but I think the part that is hard to defend is the permission, not the module.

- A customer app can only reach it if the deployment maps `/dev/crypto` into the container, and our OpenSSL has no cryptodev engine, so the app would bring its own. Possible, just not something we ship.
- Meanwhile the node is world read/write, so every non-root service on the device can open it: `aziot-keyd`, `aziot-certd`, `aziot-identityd`, `aziot-tpmd`, `aziot-edged`, `deviceupdate-agent`, `polkitd`. One of them being compromised plus this use-after-free is a root escalation.
- "Monitor for a fix" has no end date here: the CVE is in cryptodev-linux (not openssl), and the recipe already pins the newest upstream commit - 08644db02d from 2025-11-03, a Linux 6.18 build fix. There is no fix and no sign of one, so the finding comes back with every release.

Proposal: keep the module, but ship a udev rule that makes `/dev/crypto` root-only. An IoT Edge module that maps the device runs as root by default, so the use case survives, and the local escalation path from our own service accounts is gone. The moderation then reads "requires root" instead of "any local process".

If you agree I drop the cryptodev hunk from this PR - the openssl and nss cleanups are independent - and open a follow-up with the udev rule. If you would rather leave it fully as it is, that is fine too; then the entry stays EXPLOITABLE and we re-moderate it each release.

mlilien: @HarryWaschkeit I'm fine with the udev rule, a customer could override it if needed. your opinion?

mlilien:
> @mlilien one open question, not part of this PR: nss ships ~35 more tools (`certutil`, `pk12util`, `selfserv`, `tstclnt`, `ssltap`, `signtool`, ...) because the recipe packages the whole `bindir`, and networkmanager only needs the libraries.
> 
> Keep them, or restrict `FILES:${PN}` to the libraries plus `shlibsign` (needed by the nss postinst) and optionally `certutil`? Your call, I implement it either way.

restrict

conf/machine/include/phytec-imx8mm.inc
HarryWaschkeit: @mlilien @JanZachmann udev rule is fine with me, too.

JanZachmann:
@mlilien added the nspr part from the perl discussion in DtrackPolicies PR 15401: `RDEPENDS:${PN}-dev:remove = "perl"`.

Reason: perl is in the image SBOM although no perl file ships. The SBOM is written per built recipe, and nspr is the only recipe with an edge to target perl in the build graph (`networkmanager -> nss -> nspr`, then `RDEPENDS:${PN}-dev += "perl"`). We never install dev packages, so the dependency only costs build time and creates findings.

The one thing CI has to show: the perl recipe is really not built any more, and nothing else needs it - an SDK build would be the candidate.

recipes-support/nspr/nspr_%.bbappend
```diff
@@ -0,0 +1,5 @@
+# nspr-dev is the only recipe in the image build that requires target perl, and we
+# never install dev packages. Dropping the dependency keeps the perl recipe out of
+# the build and out of the SBOM, where it otherwise shows up as a shipped
+# component although no perl file is in the image.
+RDEPENDS:${PN}-dev:remove = "perl"
```
mlilien: we have a block in the distro.conf for that. (btw: to much comment)

conf/machine/include/phytec-imx8mm.inc
JanZachmann: Done - the module stays, the mode does not.

New file-only recipe `cryptodev-udev` (same pattern as `tpm-udev`) ships `/etc/udev/rules.d/10-cryptodev.rules`:

```
KERNEL=="crypto", SUBSYSTEM=="misc", MODE="0600"
```

Installed on tauril2 through `IMAGE_INSTALL:append:pn-omnect-os-image` in the machine include; the `BAD_RECOMMENDATIONS` line is gone. The low number is deliberate: udev applies rules in lexical order, so a customer rule with a higher number overrides the mode, which is the "customer could override it if needed" part.

For the moderation this means the rationale changes from "any local process can open it" to "only root can" once this lands. I adapt the entry in DtrackPolicies PR 15404 when the release with the rule is known.

recipes-support/nspr/nspr_%.bbappend
JanZachmann: Moved into the perl block of `omnect-os-distro.conf` next to the other two removals, and the bbappend is gone:

```
# nspr-dev is the only recipe pulling target perl into the build
RDEPENDS:nspr-dev:remove = "perl"
```

Comments in the openssl and nss bbappends shortened as well.

mlilien (dismissed):

recipes-support/nspr/nspr_%.bbappend
JanZachmann: Moving the "not verified here" part out of the PR description, since that belongs to this review round and not to the change itself.

No local build was done. What CI has to confirm:

- the files are gone from the image and no `installed-vs-shipped` QA warning appears for openssl or nss
- the perl recipe is no longer built and nothing else needs it (an SDK or nativesdk build would be the one to watch)
- `/dev/crypto` on tauril2 comes up as `0600 root:root`, and the smoke tests stay green

HarryWaschkeit (approved):

recipes-support/nss/nss_%.bbappend
```diff
@@ -0,0 +1,12 @@
+# nss packages the whole bindir, but the image uses none of the tools. Keep
+# shlibsign, which the postinst needs if it runs on first boot, and move the rest
+# into a package that is not installed.
+PACKAGES:append:class-target = " ${PN}-tools"
```
JanZachmann: @mlilien implemented your "restrict" here.

- `FILES:${PN}` now takes the libraries, `${sysconfdir}` and `${bindir}/shlibsign`; everything else in `${bindir}` goes to the new `nss-tools` package, which no image installs. The tools stay buildable, so a debug image can pull them in.
- `shlibsign` stays in the main package: the nss postinst calls it, and if that postinst is deferred to first boot it needs the target binary.
- `certutil` is **not** kept - nothing in the image references it (checked all scripts, units and binaries of the 6.0.2 images). Say so and I add it back.
- The single-file removals are gone; `bltest`, `dbtool`, `ecperf`, `fipstest` and `smime` are covered by the same packaging change now.
- The override is `:class-target`, so the native and nativesdk variants keep the upstream packaging and an SDK build cannot fail on unpackaged files.

openssl keeps the file removal in `do_install`: upstream splits those two paths into `openssl-misc`, and they only end up in the main package because the distro drops that package. Re-adding `openssl-misc` to `PACKAGES` would be the cleaner fix, but it needs a build to prove perl does not come back into the image, so I left it out of this PR.

CI still has to show that the tools are gone from the image, that no `installed-vs-shipped` QA warning appears for nss, and that the nss postinst still succeeds.

JanZachmann: @HarryWaschkeit your approval is on `024b4c0c`; one commit came in after it, so please have another look at `1ca19f9c`.

What changed: the nss part no longer deletes single binaries. `FILES:${PN}` now takes the libraries, `${sysconfdir}` and `${bindir}/shlibsign`, and the rest of `${bindir}` goes into a new `nss-tools` package that no image installs - that is Marcel's "restrict" from the comment above. Nothing else in the PR moved; the cryptodev udev rule you already agreed to is unchanged.

The two points worth your eyes: `shlibsign` is the only tool left in the main package (the nss postinst calls it, and it can run on first boot), and the override is `:class-target` so native and nativesdk packaging stay as upstream has them.

recipes-support/nss/nss_%.bbappend
```diff
@@ -0,0 +1,12 @@
...
+
+FILES:${PN}:class-target = " \
+    ${sysconfdir} \
+    ${bindir}/shlibsign \
+    ${libdir}/lib*.chk \
+    ${libdir}/lib*.so \
+"
+FILES:${PN}-tools = "${bindir}"
```
HarryWaschkeit: wouldn't the three bindir files in package "${PN}" also land in package "${PN}-tools" and an installation of both therefore fail?

JanZachmann: No, they cannot land in both, and nothing stays unpackaged. Checked in this release's oe-core (`meta/lib/oe/package.py`), not from memory:

- `populate_packages()` walks `PACKAGES` in order with one global `seen` list and does `if file in seen: continue` before claiming a path. So the first package whose `FILES` matches a path gets it, and no later package can take it again.
- `nss` comes before `nss-tools` here, because the bbappend uses `PACKAGES:append`. So `./usr/bin/shlibsign` is claimed by `nss`.
- `files_from_filevars()` expands a directory entry to its children: `if cpath.isdir(f): files += [os.path.join(f, x) for x in os.listdir(f)]`. So `FILES:${PN}-tools = "${bindir}"` becomes `./usr/bin` plus every file in it. `./usr/bin` itself is already in `seen` (it was created while packaging `shlibsign`) and gets skipped; the remaining binaries are not in `seen`, so they go to `nss-tools`. `shlibsign` is skipped there for the same reason.
- The `/usr/bin` directory node exists in both packages - `mkdir_recurse()` creates it per package. Shared directories are not a file conflict for ipk/rpm, only identical files would be.
- Consequence for the QA check: every installed path ends up in `seen`, so no `installed-vs-shipped`.

The same recipe already relies on that rule in the other direction: `FILES:${PN}-smime = "${bindir}/smime"` next to `FILES:${PN} = "... ${bindir} ..."`, with `nss-smime` prepended, so it wins for that one file.

Small correction on the count: `FILES:${PN}` has one `${bindir}` entry (`shlibsign`); the other three entries are `${sysconfdir}` and the two `${libdir}/lib*` globs.

One caveat this does depend on: `nss-tools` must stay **after** `${PN}` in `PACKAGES`. If someone changes it to `PACKAGES =+`, the whole `${bindir}` including `shlibsign` moves into `nss-tools` and the main package loses it silently. If you would rather not have the intent depend on ordering plus the directory-expansion rule, I can write `FILES:${PN}-tools = "${bindir}/*"` - same result, but the file says "the files in bindir", not "the directory".

Limitation: this is from reading the packaging code, not from a build - I have no build environment here, so CI still has to show the split and that no QA warning appears.

mlilien (approved):

PR #689, opened 2026-08-21
systemd: build without TPM2 support

JanZachmann:
**Summary**

Build systemd without its TPM2 feature set (`-Dtpm2=disabled`). That drops `systemd-tpm2-setup.service` and `-early`, the NvPCR initialization and the `systemd-pcr*` units.

**Reason**

`systemd-tpm2-setup` seals an NvPCR anchor secret into the TPM on every boot. The seal is dictionary attack protected, so it fails while the TPM's failed-tries counter sits at `maxTries`: the unit fails, the system state turns `degraded`, and an update under validation is rolled back although the update is fine. The counter climbs on its own, because every reset without an orderly `TPM2_Shutdown` adds one and nothing ever resets it.

Nothing uses the feature: there are no encrypted volumes, no TPM sealed credentials, and no reference to NvPCRs in this layer. The TPM stays available to the identity service, which links `libtss2` directly, and to `tpm2-tools`. Images built before the feature was turned on report `-TPM2` and ship no such unit.

### DIFF
--- a/recipes-core/systemd/systemd_%.bbappend
+++ b/recipes-core/systemd/systemd_%.bbappend
@@ -8,6 +8,11 @@ SRC_URI += "\
 
 RDEPENDS:${PN} += "bash"
 
+# systemd's TPM2 features are unused (no encrypted volumes, no sealed
+# credentials) and they pull in systemd-tpm2-setup, which fails once the TPM's
+# dictionary attack counter is exhausted and then leaves the system degraded.
+PACKAGECONFIG:remove = "tpm2"
+
 # enable bash-completion
 bashcompletiondir = "${datadir}/bash-completion/completions"
 


HarryWaschkeit (approved):

JanZachmann:
@mlilien can you have a look again? The PR description was updated to our latest discussion.

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -8,6 +8,11 @@ SRC_URI += "\
 
 RDEPENDS:${PN} += "bash"
 
+# systemd's TPM2 features are unused (no encrypted volumes, no sealed
+# credentials) and they pull in systemd-tpm2-setup, which fails once the TPM's
```
mlilien: do we want to document that they pull in systemd-tpm2-setup for efi devices only?

JanZachmann: Added in 4d267ed. The units are installed on every machine, but they carry ConditionSecurity=tpm2, which needs TPM2_SUPPORT_FIRMWARE ("firmware reports TPM2 was used"), so only EFI machines actually run them. Measured on 6.0.3 wrynose with the same TPM DA parameters on all boards: the EFI board goes degraded under lockout, the u-boot boards stay running.

Verification for this round: the change is comment-only, no build impact. The behaviour itself was verified on three boards (i.MX8MM u-boot, rpi4 + SLB9670 u-boot, genericx86-64 EFI).

mlilien (approved):

PR #690, opened 2026-08-21
chore: drop unused image files and the perl build dependency, restrict /dev/crypto

JanZachmann:
## Summary

- **openssl** bbappend: delete `c_rehash` and `${libdir}/ssl-3/misc`
- **nss** bbappend: restrict `FILES:${PN}` to the libraries, the config and `shlibsign`, and move the rest of `${bindir}` into a new `nss-tools` package that no image installs
- **omnect-os-distro.conf**: `RDEPENDS:nspr-dev:remove = "perl"`
- **cryptodev-udev**: new file-only recipe shipping `10-cryptodev.rules`, which gives `/dev/crypto` mode `0600`; installed on tauril2 through the machine include. The module itself stays in the image.

## Reason

`omnect-os-distro.conf` already drops `openssl-misc` and `nss-smime` from `PACKAGES` to keep perl out of the image, but `PACKAGES:remove` only removes the package - the files move into the respective main package and still ship, as perl scripts in an image without a perl interpreter.

nss is the bigger case: the recipe has `FILES:${PN} = "... ${bindir} ..."`, so **49 tools** ship (`certutil`, `pk12util`, `selfserv`, `tstclnt`, `ssltap`, `signtool`, the test binaries, ...) although nothing in the image references any of them - networkmanager pulls nss in for the libraries only. Restricting the package and collecting the tools in `nss-tools` keeps them buildable for a debug image and takes them out of the release image, including `smime`, so no single-file removals are needed there. `shlibsign` stays in the main package because the nss postinst calls it, and that postinst can end up running on first boot. openssl keeps the file removal: upstream splits those two paths into `openssl-misc` and the main package claims them only because the distro drops that package.

perl itself shows up in the image SBOM although no perl file is in the image: the SBOM is written per built recipe, and the only recipe pulling target perl into the build is nspr (`RDEPENDS:${PN}-dev += "perl"`, nspr comes in via `networkmanager -> nss -> nspr`). Dev packages are never installed, so dropping that dependency changes nothing in the image while taking perl out of the build and out of the SBOM - four high/critical findings less to moderate per release.

On tauril2 the imx BSP installs the cryptodev module, it is loaded at boot, and the driver creates `/dev/crypto` world read/write, while upstream has no fix for CVE-2026-28529 (use-after-free in `get_userbuf()`, local privilege escalation). The module stays available for applications; the udev rule takes away the world read/write default, and its low number lets a customer rule override the mode.

No other BSP ships cryptodev, and no script, unit or binary in the 6.0.2 images references the removed files or `/dev/crypto`.

### DIFF
--- a/conf/machine/include/phytec-imx8mm.inc
+++ b/conf/machine/include/phytec-imx8mm.inc
@@ -23,6 +23,9 @@ IMAGE_BOOT_FILES:remove:pn-omnect-os-image = " \
 
 IMAGE_INSTALL:append:pn-omnect-os-image = " kernel-devicetree"
 
+# the imx BSP installs the cryptodev module; restrict its device node to root
+IMAGE_INSTALL:append:pn-omnect-os-image = " cryptodev-udev"
+
 # initramfs type
 OMNECT_INITRAMFS_FSTYPE = "cpio.gz.u-boot"
 
--- a/recipes-support/nss/nss_%.bbappend
+++ b/recipes-support/nss/nss_%.bbappend
@@ -1,0 +1,12 @@
+# nss packages the whole bindir, but the image uses none of the tools. Keep
+# shlibsign, which the postinst needs if it runs on first boot, and move the rest
+# into a package that is not installed.
+PACKAGES:append:class-target = " ${PN}-tools"
+
+FILES:${PN}:class-target = " \
+    ${sysconfdir} \
+    ${bindir}/shlibsign \
+    ${libdir}/lib*.chk \
+    ${libdir}/lib*.so \
+"
+FILES:${PN}-tools = "${bindir}"
--- a/recipes-support/nss/nss_%.bbappend
+++ b/recipes-support/nss/nss_%.bbappend
@@ -1,6 +1,5 @@
 # nss packages the whole bindir, but the image uses none of the tools. Keep
-# shlibsign, which the postinst needs if it runs on first boot, and move the rest
-# into a package that is not installed.
+# shlibsign, which the postinst needs if it runs on first boot.
 PACKAGES:append:class-target = " ${PN}-tools"
 
 FILES:${PN}:class-target = " \
@@ -9,4 +8,6 @@ FILES:${PN}:class-target = " \
     ${libdir}/lib*.chk \
     ${libdir}/lib*.so \
 "
+# PACKAGES order decides who claims a file first: nss takes shlibsign, nss-tools
+# the rest of bindir. No file ends up in both.
 FILES:${PN}-tools = "${bindir}"
--- a/conf/distro/include/omnect-os-distro.conf
+++ b/conf/distro/include/omnect-os-distro.conf
@@ -163,6 +163,8 @@ PACKAGECONFIG:pn-opkg-utils = "update-alternatives"
 # get rid of perl as runtime dependency in release image
 PACKAGES:remove:pn-nss = "nss-smime"
 PACKAGES:remove:pn-openssl = "openssl-misc"
+# nspr-dev is the only recipe pulling target perl into the build
+RDEPENDS:nspr-dev:remove = "perl"
 PACKAGECONFIG:remove:pn-apparmor = "perl"
 DEPENDS:remove:pn-apparmor = "perl"
 # apparmor inherits cpan, which exports PERLCONFIGTARGET=yes and makes the native
--- a/recipes-connectivity/openssl/openssl_%.bbappend
+++ b/recipes-connectivity/openssl/openssl_%.bbappend
@@ -1,0 +1,6 @@
+# openssl-misc is dropped from PACKAGES, which leaves its perl scripts in the
+# main package - remove the files as well.
+do_install:append:class-target() {
+    rm -f ${D}${bindir}/c_rehash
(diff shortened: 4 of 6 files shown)


conf/machine/include/phytec-imx8mm.inc
```diff
@@ -23,6 +23,9 @@ IMAGE_BOOT_FILES:remove:pn-omnect-os-image = " \
 
 IMAGE_INSTALL:append:pn-omnect-os-image = " kernel-devicetree"
 
+# the imx BSP installs the cryptodev module; restrict its device node to root
+IMAGE_INSTALL:append:pn-omnect-os-image = " cryptodev-udev"
```
mlilien: i should've asked this in the previous PR: is it really only relevant for phytec devices? is the cryptodev device only available there?

JanZachmann: Today it is only tauril2, checked at SBOM level rather than by reading recipes: Dependency-Track lists CVE-2026-28529 (component `cryptodev-linux`) for `omnect-os-gateway-tauril2` and `omnect-os-gateway-devel-tauril2` only, on 5.0.17 and 6.0.2 - rpi4, genericx86-64 and dehndetect have no cryptodev component at all.

Where it comes from: the module is oe-core's `cryptodev-module`, and the NXP BSP layer is what carries it into the build - `meta-imx/meta-imx-bsp/recipes-kernel/cryptodev/cryptodev-linux_1.14.bbappend` (it adds an RSA patch). meta-imx is only in the tauril2 build. Note it is not "every i.MX board": dehndetect (tx6s, i.MX6) does not ship cryptodev either, so the trigger is the meta-imx BSP, not the SoC family.

That said, your question points at the weak spot of putting it in the machine include: the next machine that pulls in meta-imx would silently ship `/dev/crypto` as 0666 again, and nobody would notice until the next CVE round. The rule itself is machine-independent - `KERNEL=="crypto", SUBSYSTEM=="misc", MODE="0600"` never matches on an image without the module - so installing it unconditionally costs three lines in the rootfs and has no runtime effect anywhere else.

So: keep it narrow where the module actually is, or move the `IMAGE_INSTALL` line from `phytec-imx8mm.inc` to `omnect-os-distro.conf` so it cannot be forgotten? I would move it, for that reason. Say which you prefer and I push it.

mlilien (approved):

recipes-support/nss/nss_%.bbappend
```diff
@@ -0,0 +1,12 @@
...
+# into a package that is not installed.
+PACKAGES:append:class-target = " ${PN}-tools"
+
+FILES:${PN}:class-target = " \
+    ${sysconfdir} \
+    ${bindir}/shlibsign \
+    ${libdir}/lib*.chk \
+    ${libdir}/lib*.so \
+"
+FILES:${PN}-tools = "${bindir}"
```
HarryWaschkeit: won't that place the three listed binaries in both packages and lead to an installation failure due to duplicated files if also both are to be part of thi image?

JanZachmann: No, and this is the same question you raised on #688 - answer repeated here so this PR stands on its own. Checked in this release's oe-core (`meta/lib/oe/package.py`), not from memory:

- `populate_packages()` walks `PACKAGES` in order with one global `seen` list and does `if file in seen: continue` before claiming a path. First matching package wins; no later package can take the same path.
- `nss` comes before `nss-tools` (the bbappend uses `PACKAGES:append`), so `${bindir}/shlibsign` is claimed by `nss`.
- When `nss-tools` is processed, `files_from_filevars()` expands the directory entry to its children (`if cpath.isdir(f): files += [os.path.join(f, x) for x in os.listdir(f)]`). `./usr/bin` itself is already in `seen` (it was created while packaging `shlibsign`) and is skipped, `shlibsign` is skipped for the same reason, and the remaining binaries go to `nss-tools`.
- So no file is in both packages, and no installed path stays unpackaged either - hence no `installed-vs-shipped`.
- The `/usr/bin` directory node exists in both packages; `mkdir_recurse()` creates it per package. Shared directories are not a conflict for ipk/rpm, only identical files would be.
- The recipe already relies on the same rule in the other direction: `FILES:${PN}-smime = "${bindir}/smime"` next to `FILES:${PN} = "... ${bindir} ..."`, with `nss-smime` prepended.

One correction on the count: `FILES:${PN}` has a single `${bindir}` entry (`shlibsign`); the other three entries are `${sysconfdir}` and the two `${libdir}/lib*` globs.

Since the line has now prompted the question twice, `1ee9be3` writes the rule into the file:

```
# PACKAGES order decides who claims a file first: nss takes shlibsign, nss-tools
# the rest of bindir. No file ends up in both.
FILES:${PN}-tools = "${bindir}"
```

The only alternative that removes the implicit ordering would be listing all ~49 tool names in `FILES:${PN}-tools` and putting that package before `${PN}` (the upstream `nss-smime` pattern). I would not do that for 49 names, but say so if you disagree.

Limitation unchanged: this is from reading the packaging code, not from a build - CI still has to show the split and that no QA warning appears.

@mlilien your approval was on `86113fc7` and `main` requires approval of the last push, so the comment commit above invalidated it - re-requesting.

HarryWaschkeit (approved):

